Issue #12, opened 2022-12-20
IndexingError: Unalignable boolean Series provided as indexer (index of the boolean Series and of the indexed object do not match).


whiteorchid:
Dear Authors,

Thanks a lot for the great tool. When using the own data and transform the data format similar to the tutorial data. The  dds.deseq2 raise the error. 

```
`Fitting size factors...
... done in 0.03 seconds.

Fitting dispersions...
... done in 10.72 seconds.

Fitting dispersion trend curve...
... done in 9.18 seconds.

---------------------------------------------------------------------------
IndexingError                             Traceback (most recent call last)
Cell In[93], line 2
      1 # Then, run DESeq2 on it
----> 2 dds.deseq2()

File ~/miniconda3/envs/pydeseq2/lib/python3.10/site-packages/pydeseq2/DeseqDataSet.py:191, in DeseqDataSet.deseq2(self)
    189 self.fit_dispersion_trend()
    190 # Compute prior dispersion variance
--> 191 self.fit_dispersion_prior()
    192 # Refit genewise dispersions a posteriori (shrinks estimates towards trend curve)
    193 self.fit_MAP_dispersions()

File ~/miniconda3/envs/pydeseq2/lib/python3.10/site-packages/pydeseq2/DeseqDataSet.py:365, in DeseqDataSet.fit_dispersion_prior(self)
    361 # Compute squared log-residuals and prior variance based on genes whose
    362 # dispersions are above 100 * min_disp. This is to reproduce DESeq2's behaviour.
    363 above_min_disp = self.genewise_dispersions >= (100 * self.min_disp)
    364 self._squared_logres = np.abs(
--> 365     disp_residuals[above_min_disp]
    366 ).median() ** 2 / norm.ppf(0.75)
    367 self.prior_disp_var = np.maximum(
    368     self._squared_logres - polygamma(1, (m - p) / 2), 0.25
    369 )

File ~/miniconda3/envs/pydeseq2/lib/python3.10/site-packages/pandas/core/series.py:1003, in Series.__getitem__(self, key)
   1000     key = list(key)
   1002 if com.is_bool_indexer(key):
-> 1003     key = check_bool_indexer(self.index, key)
   1004     key = np.asarray(key, dtype=bool)
   1005     return self._get_values(key)

File ~/miniconda3/envs/pydeseq2/lib/python3.10/site-packages/pandas/core/indexing.py:2552, in check_bool_indexer(index, key)
[...]

PR #14, opened 2022-12-20
Handling all zero genes

BorisMuzellec:
Handle genes that only have 0 counts by excluding them from fits and returning NaNs, as DESeq2 does.


mandreux-owkin (commented):
The changes seem legit to me, but what do you think adding a test with some input data with real zeros, that would not pass with the previous version, but now passes with this patch? Linked to #12 for reference.

mandreux-owkin (approved):

BorisMuzellec:
Thanks, will open an issue to add unit tests for inputs with zeros / negative values / NaNs

PR #26, opened 2022-12-28
CI: implement tests for design factors

BorisMuzellec:
This PR implements tests on the design factor in the clinical data. More precisely:

- The `DeseqDataset` class now tests whether the design factor contains NaN values and checks the design factors takes exactly 2 values. 
- Pytests now test that an error is thrown in case a `DeseqDataset` is initialised with NaNs, or not exactly two values in the design factor.

DO NOT MERGE: there are a few tests that remain to be implemented

### DIFF
--- a/pydeseq2/DeseqDataSet.py
+++ b/pydeseq2/DeseqDataSet.py
@@ -47,13 +47,13 @@ class DeseqDataSet:
         Name of the column of clinical to be used as a design variable.
         (default: 'high_grade').
 
-    min_mu : float
-        Threshold for mean estimates. (default: 0.5).
-
     reference_level : str
         The factor to use as a reference. Must be one of the values taken by the design.
         If None, the reference will be chosen alphabetically (last in order).
         (default: None).
+
+    min_mu : float
+        Threshold for mean estimates. (default: 0.5).
 
     min_disp : float
         Lower threshold for dispersion parameters. (default: 1e-8).
--- a/pydeseq2/DeseqDataSet.py
+++ b/pydeseq2/DeseqDataSet.py
@@ -177,6 +177,10 @@ class DeseqDataSet:
 
         # Import clinical data and convert design_column to string
         self.clinical = clinical
+        if not self.counts.index.identical(self.clinical.index):
+            raise ValueError(
+                "The count matrix and clinical data should have the same index."
+            )
         self.design_factor = design_factor
         if self.clinical[self.design_factor].isna().any():
             raise ValueError("NaNs are not allowed in the design factor.")
--- a/pydeseq2/DeseqDataSet.py
+++ b/pydeseq2/DeseqDataSet.py
@@ -165,10 +165,13 @@ class DeseqDataSet:
 
         # Test counts before going further
         test_valid_counts(counts)
-
         self.counts = counts
+
+        # Import clinical data and convert design_column to string
         self.clinical = clinical
         self.design_factor = design_factor
+        self.clinical[self.design_factor] = self.clinical[self.design_factor].astype(str)
+
         # Build the design matrix (splits the dataset in two cohorts)
         self.design_matrix = build_design_matrix(
             self.clinical, design_factor, expanded=False, intercept=True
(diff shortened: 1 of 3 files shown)


pydeseq2/DeseqDataSet.py
```diff
@@ -46,6 +46,10 @@ class DeseqDataSet:
     design_factor : str, default='high_grade'
         Name of the column of clinical to be used as a design variable.
 
+    reference_level : str, default=None
+        The factor to use as a reference. Must be one of the values taken by the design.
+        If None, the reference will be chosen alphabetically (last in order).
+
```
arthurPignetOwkin, line 52:
```suggestion
    reference_level : str
        The factor to use as a reference. Must be one of the values taken by the design.
        If None, the reference will be chosen alphabetically (last in order). (default: None).

```

pydeseq2/utils.py
```diff
@@ -170,14 +170,33 @@ def build_design_matrix(
...
+    if design_matrix.shape[-1] == 1:
+        raise ValueError(
+            f"The design factor takes only one value: "
+            f"{design_matrix.columns.values.tolist()}, but two are necessary."
+        )
+    elif design_matrix.shape[-1] > 2:
+        raise ValueError(
+            f"The design factor takes more than two values: "
+            f"{design_matrix.columns.values.tolist()}, but should have exactly two."
+        )
```
arthurPignetOwkin, line 185: Could the tests be merged ? using != 2, f'the design factor takes {len(design_matrix.shape[-1])} values: {design_matrix.columns.values.tolist()}, but should have exactly 2'

pydeseq2/utils.py
```diff
@@ -170,14 +170,33 @@ def build_design_matrix(
...
-        design_matrix.insert(0, ref, ref_level)
+    if ref is not None:  # Put reference level last
+        try:
+            ref_level = design_matrix.pop(ref)
+        except KeyError as e:
+            print(
+                "The reference level must correspond to"
+                " one of the design factor values."
+            )
+            raise e
```
arthurPignetOwkin, line 198: I am more used to 

```python
except KeyError as e:
            raise e( "The reference level must correspond to"
            +  " one of the design factor values.")
```
But I don't know if one or the other is best. What do you think ?

tests/test_edge_cases.py
```diff
@@ -89,3 +90,32 @@ def test_non_negative_counts():
 
     with pytest.raises(ValueError):
         DeseqDataSet(counts_df, clinical_df, design_factor="condition")
+
+
```
arthurPignetOwkin, line 94: Could you also test that a key error is raised when the ref level isn't in the design factor values ? I think it will be the most common error

pydeseq2/DeseqDataSet.py
arthurPignetOwkin: Change the default to numpydoc api

arthurPignetOwkin (changes requested):

pydeseq2/utils.py
BorisMuzellec: I don't know, it seems more informative this way? Although it would save a couple of code lines indeed

pydeseq2/utils.py
BorisMuzellec: Indeed

pydeseq2/utils.py
arthurPignetOwkin: From my point of view printing "but two are necessary." is a less informative compare to "but should have exactly two.", so I would copy-past the sentence, and thus the tests. Refactoring would make the code clearer I think. But it is not important so I would approve even without this change

tests/test_edge_cases.py
BorisMuzellec: Done in [0bdcd67](https://github.com/owkin/PyDESeq2/pull/26/commits/0bdcd67dcb46bd35ff7e27fcb3862c09abe2fbc1)

pydeseq2/utils.py
arthurPignetOwkin: you are right, 
The syntax should be 
```python
try:
   pass
except KeyError as e:
   raise KeyError('mon text') from e
``` 
Or create a custom exception. I wonder if the print in an except goes to sys.stdout or sys.stderr

arthurPignetOwkin (approved):
LGTM!

PR #30, opened 2023-01-02
Index sorting

BorisMuzellec:
This PR fixes an issue that appears when the count matrix and the clinical data provided to a `DeseqDataSet` have identical sets of samples, but in a different order, by sorting the indices at initialization.

### DIFF
--- a/pydeseq2/DeseqDataSet.py
+++ b/pydeseq2/DeseqDataSet.py
@@ -174,15 +174,19 @@ class DeseqDataSet:
         # Test counts before going further
         test_valid_counts(counts)
         self.counts = counts
-        self.counts.sort_index(inplace=True)
 
         # Import clinical data and convert design_column to string
-        self.clinical = clinical
-        self.clinical.sort_index(inplace=True)
-        if not self.counts.index.identical(self.clinical.index):
-            raise ValueError(
-                "The count matrix and clinical data should have the same index."
+        try:
+            self.clinical = clinical.loc[self.counts.index]
+        except KeyError as e:
+            print(
+                "The count matrix and clinical data "
+                "should contain the same sample indexes."
             )
+            raise e
+        assert len(self.clinical) == len(
+            clinical
+        ), "The count matrix and clinical data should contain the same sample indexes."
         self.design_factor = design_factor
         if self.clinical[self.design_factor].isna().any():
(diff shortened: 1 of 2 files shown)


pydeseq2/DeseqDataSet.py
```diff
@@ -176,11 +176,17 @@ def __init__(
...
-                "The count matrix and clinical data should have the same index."
+        try:
+            self.clinical = clinical.loc[self.counts.index]
+        except KeyError as e:
+            print(
+                "The count matrix and clinical data "
+                "should contain the same sample indexes."
             )
+            raise e
+        assert len(self.clinical) == len(
```
arthurPignetOwkin, line 187: Why do you raise an assertion error here and not a KeyError again ? Unless I misunderstood the message is the same no ?

BorisMuzellec: I don't know, maybe I could raise a KeyError again... The reason behind the assertion is that `clinical.loc[self.counts.index]` will work when the index of `counts` is a subset of the index of `clinical`, but will drop rows from `clinical`.

BorisMuzellec: Simplified this part in [2d7ce6e](https://github.com/owkin/PyDESeq2/pull/30/commits/2d7ce6eb4e3246a12ebab69a2a9a3c0b95c147fc)

maikia (approved):
Thanks for your contribution @BorisMuzellec  ;-p 

This is already great, if you want to make it even better you can also check if the correct error message is raised with the error to ensure we are not just raising any keyerror but feel free to merge as is

tests/test_edge_cases.py
```diff
@@ -141,7 +141,7 @@ def test_indexes():
     )
     clinical_df = pd.DataFrame({"condition": [0, 1]}, index=["sample01", "sample02"])
 
-    with pytest.raises(ValueError):
+    with pytest.raises((KeyError, AssertionError)):
```
BorisMuzellec, line 144: Shouldn't raise an `AssertionError` anymore.

```suggestion
    with pytest.raises(KeyError):
```

BorisMuzellec:
Thanks @maikia and @arthurPignetOwkin ! I'll merge then

PR #31, opened 2023-01-05
DOC rearrange home page

maikia:
- Changed the welcoming message (I think it wasn't gramatically correct)

- Rearranged the text so that now instead of having all the explanations on the home page there are tabs so it looks like that:

image


BorisMuzellec (approved):
LGTM, thanks!

PR #32, opened 2023-01-05
Paired multi-factor analysis

BorisMuzellec:
This PR aims to implement paired multi-factor analysis


pydeseq2/DeseqDataSet.py
```diff
@@ -181,15 +182,22 @@ def __init__(
...
+        self.clinical[self.design_factors] = self.clinical[self.design_factors].astype(
+            str
+        )
 
-        # Build the design matrix (splits the dataset in two cohorts)
+        # Build the design matrix (splits the dataset in cohorts)
         self.design_matrix = build_design_matrix(
             self.clinical,
-            design_factor,
+            design_factors,
```
maikia, line 200: just a minor comment: I would suggest to add the argument names which adds for the clarity of read (not required for merge though)

pydeseq2/DeseqDataSet.py
```diff
@@ -248,6 +256,8 @@ def fit_genewise_dispersions(self):
         Fits a negative binomial per gene, independently.
         """
 
+        p = self.design_matrix.shape[-1]
```
maikia, line 259: I am never a fun of single letters as variable names therefore I would suggest to make it more meaningful

pydeseq2/DeseqDataSet.py
```diff
@@ -266,22 +276,47 @@ def fit_genewise_dispersions(self):
         rough_disps = self._rough_dispersions.loc[self.non_zero_genes].values
         size_factors = self.size_factors.values
 
-        with parallel_backend("loky", inner_max_num_threads=1):
-            mu_hat_ = np.array(
-                Parallel(
+        # mu_hat is initialized differently depending on the number of different factor
```
maikia, line 279: do you think it might be useful to through a warning or otherwise give the comment to the user on which fit was used? Or at least add it to the docstring so it is easier to find? (please ignore if you don't think it is useful information to the user)

pydeseq2/DeseqStats.py
```diff
@@ -28,8 +29,15 @@ class DeseqStats:
     dds : DeseqDataSet
         DeseqDataSet for which dispersion and LFCs were already estimated.
 
+    contrast : list[str] or None
```
maikia, line 32: from this description I am not sure what contrast actually is (just the format of it) and when to use it instead of None. Could you add a short description?

pydeseq2/DeseqStats.py
```diff
@@ -108,6 +120,19 @@ def __init__(
         ), "Please provide a fitted DeseqDataSet by first running the `deseq2` method."
 
         self.dds = dds
+
+        # Build contrast if None
+        if contrast is not None:
+            # TODO : tests on the contrast
+            self.contrast = contrast
```
maikia, line 127: maybe check if the contrast is indeed given as required? ie three strings in a list and through an error otherwise

pydeseq2/DeseqStats.py
```diff
@@ -151,15 +172,100 @@ def summary(self):
...
+        # Compute means according to the LFC model.
+        mu = (
+            np.exp(self.dds.design_matrix @ self.LFCs.T)
+            .multiply(self.dds.size_factors, 0)
+            .values
+        )
+
+        # Set regularization factors.
+        if self.prior_disp_var is not None:
+            D = np.diag(1 / self.prior_disp_var**2)
```
maikia, line 233: same as above, could you change D for more informative name?

pydeseq2/DeseqStats.py
```diff
@@ -202,6 +308,8 @@ def lfc_shrink(self):
                     offset,
                     prior_no_shrink_scale,
                     prior_scale,
+                    "L-BFGS-B",
```
maikia, line 311: maybe add a comment that it's an optimizer

pydeseq2/utils.py
```diff
@@ -144,15 +144,16 @@ def build_design_matrix(
...
 
     Parameters
     ----------
     clinical_df : pandas.DataFrame
         DataFrame containing clinical information.
         Must be indexed by sample barcodes, and contain a "high_grade" column.
 
-    design : str
-        Name of the column of clinical_df to be used as a design_matrix variable.
+    design : str or list
```
maikia, line 155: list of strings? 
```suggestion
    design : str or list[str]
```

pydeseq2/utils.py
```diff
@@ -675,6 +683,7 @@ def trimmed_variance(x, trim=0.125, axis=1):
     return 1.51 * trimmed_mean(sqerror, trim=trim, axis=axis)
 
 
+# TODO : bug here in multi-factor?
```
maikia, line 686: can you give more info for the future dev trying to fix it? when exactly does it happen, what kind of bug?

pydeseq2/utils.py
```diff
@@ -941,18 +948,25 @@ def nbinomGLM(
     converged: bool
         Whether L-BFGS-B converged.
     """
-    beta_init = np.array([0.1, -0.1])
-    no_shrink_index = 1 - shrink_index
+
+    p = design_matrix.shape[-1]
```
maikia, line 952: here again I would suggest changing the name, up to teste

pydeseq2/utils.py
maikia: you can also add it to the list of issues after merging with exact explanation, with the code snippet on how to get it and what you expect instead

pydeseq2/utils.py
BorisMuzellec: Thanks, done in [e536451](https://github.com/owkin/PyDESeq2/pull/32/commits/e5364518043f4edf9dccaf314ad8020bc7ba2b87)

pydeseq2/utils.py
BorisMuzellec: This is a note I left for myself which I meant to delete later on. Thanks for catching it!

pydeseq2/DeseqStats.py
BorisMuzellec: Thanks, done in [e536451](https://github.com/owkin/PyDESeq2/pull/32/commits/e5364518043f4edf9dccaf314ad8020bc7ba2b87)

pydeseq2/DeseqDataSet.py
BorisMuzellec: Good idea, changed it to `num_vars` in [e536451](https://github.com/owkin/PyDESeq2/pull/32/commits/e5364518043f4edf9dccaf314ad8020bc7ba2b87)

pydeseq2/DeseqDataSet.py
BorisMuzellec: Hi Maria, not sure what you mean, do you mean we should write `design=design_factors` ?

pydeseq2/DeseqDataSet.py
BorisMuzellec: That's a good question. Personally I don't think a warning is necessary, as the `_mu_hat` attribute is not meant to be accessed by the user (it's an intermediate value). DESeq2 does this silently too. If that's OK for you, I'll leave this as is for now.

pydeseq2/DeseqStats.py
BorisMuzellec: I've expanded the description of this argument : [8c2197a](https://github.com/owkin/PyDESeq2/pull/32/commits/8c2197a349d45c5a089b23981c58e0f270c3c66a)

Let me know if this helps, or it is still a bit obscure

pydeseq2/DeseqStats.py
BorisMuzellec: Added tests in [39f4211](https://github.com/owkin/PyDESeq2/pull/32/commits/39f4211014399175ce08cc18e24141ffd5f08936) and [b0e35f5](https://github.com/owkin/PyDESeq2/pull/32/commits/b0e35f50c69f4c625a67896f4a9837972fc59fe9)

pydeseq2/DeseqStats.py
BorisMuzellec: I've added the argument name ("`optimizer`") in [3c0ded6](https://github.com/owkin/PyDESeq2/pull/32/commits/3c0ded613d38aacac433983c9871e3f3210c37b5)

pydeseq2/DeseqDataSet.py
BorisMuzellec: I've named all arguments in [3c0ded6](https://github.com/owkin/PyDESeq2/pull/32/commits/3c0ded613d38aacac433983c9871e3f3210c37b5)

BorisMuzellec:
Thanks a lot @maikia for your review, it was super helpful :).
I've pushed a few commits based on your suggestions, could you have a look at them when you have time? Thanks!

pydeseq2/DeseqDataSet.py
maikia: yes, that's what I meant. Thanks :-)

pydeseq2/DeseqStats.py
maikia: yes, thanks!

PR #42, opened 2023-01-11
DOC autosummary for api

maikia:
Adds the autosummary for the api: meaning the table with short description for docstrings

now it looks like this:
image

and this:
image

### DIFF
--- a/docs/source/api/DeseqDataSet.rst
+++ b/docs/source/api/DeseqDataSet.rst
@@ -1,8 +1,0 @@
-pydeseq2.DeseqDataSet module
-============================
-
-.. automodule:: pydeseq2.DeseqDataSet
-   :members:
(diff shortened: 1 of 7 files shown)


BorisMuzellec (changes requested):
Thanks a lot @maikia, it looks amazing!

Just a small suggestion: would it be possible to list the class attributes of `DeseqDataSet` and `DeseqStats` the same way the methods are listed? (like in e.g. [squidpy](https://squidpy.readthedocs.io/en/stable/classes/squidpy.im.ImageContainer.html))

Would it also be possible not to list the `__init__` method in the doc (if you think this is a good idea)? It seems like it doesn't add much information compared to the description of the class, which already lists the arguments.

BorisMuzellec (approved):

Issue #43, opened 2023-01-11
Issue with Cooks Refitting if Less Than Three Replicates


LawrenceGordon:
I am working with a small dataset (two unaffected, two affected) samples. I have been running into an error in the utils function `robust_method_of_moments_disp()` which is called in `DeseqDataSet.calculate_cooks()`. The core issue I've found is in the `trimmed_variance()` function:
```
def trimmed_variance(x, trim=0.125, axis=1):

    rm = trimmed_mean(x, trim=trim, axis=axis)
    sqerror = (x - rm) ** 2
    # scale due to trimming of large squares
    return 1.51 * trimmed_mean(sqerror, trim=trim, axis=axis)
```
The returned value for `rm` is in the shape (n_samples, ) while the shape of `x` is (n_sample, n_genes). This leads to an error like follows:
```
ValueError
Unable to coerce to Series, length must be n_genes: given n_samples
```
I was able to bypass the error by using the `trimmed_cell_variance()` function instead, however a stronger fix to the issue is outside of my abilities, as I'm unsure the actual desired shape of `sqerror`.

BorisMuzellec:
Thanks for reporting this bug @LawrenceGordon.
I think I've identified where it comes from, I've opened a PR (#44) to fix it. I'll let you know when it's merged :).

BorisMuzellec:
I've merged a PR to fix this issue. It'll be part of the next PyPI release.

PR #44, opened 2023-01-12
Fix: calculate Cooks distances with few samples 

BorisMuzellec:
This PR aims to fix the bug described in Issue #43.

The issue seemed to be that the summation was done on the wrong axis in the `trimmed_variance` function, called through `robust_method_of_moments_disp` in the  `calculate_cooks` method of the `DeseqDataSet` class.

### DIFF
--- a/pydeseq2/utils.py
+++ b/pydeseq2/utils.py
@@ -672,7 +672,7 @@ def trimmed_cell_variance(counts, cells):
     return varEst.max(axis=1)
 
 
-def trimmed_variance(x, trim=0.125, axis=1):
+def trimmed_variance(x, trim=0.125, axis=0):
     """Return trimmed variance.
 
      Compute the variance after trimming data of its smallest and largest quantiles.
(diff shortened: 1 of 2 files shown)


mandreux-owkin (commented):
Very high level review, but is there any way to add a test to check the behaviour that lead to this bug? This way, we make sure the codebase does not forget about it. Also, it's a good way to ensure this PR really fixes the related issue.

BorisMuzellec:
> Very high level review, but is there any way to add a unit test to check the behaviour that lead to this bug? 

I guess we could add a test where the pipeline is run on a dataset with few samples, as the piece of code responsible for the bug only is run only when no cohort has 3 or more samples. Not sure what to test for though, would error-free execution be enough?

mandreux-owkin:
>  Not sure what to test for though, would error-free execution be enough?

Yes I think that simply adding this corner case and ensuring it runs without error is a good starting point!

mandreux-owkin (approved):
LGTM, thanks @BorisMuzellec

PR #54, opened 2023-01-16
BENCH remove support for TCGA data

BorisMuzellec:
This PR removes any type of support for TCGA data (in the `load_example_data` function and `notebooks` directory in particular).

This is because we do not provide the data ourself, nor do we provide a script or simple instructions to obtain this data.


maikia (approved):
LGTM, thanks for your work @BorisMuzellec 
feel free to merge when CI passes the tests

PR #58, opened 2023-01-23
DOC Convert step-by-step notebook to example

maikia:
Converts second notebook to example.
This PR is only to convert the notebook so that binder and notebooks can be removed from the github repo in the next PR.
This PR is not to improve the example which will be done in follow-up PR


BorisMuzellec (approved):
LGTM, thanks Maria!

PR #59, opened 2023-01-23
DOC: refactoring

BorisMuzellec:
This PR is a series of small improvements of the documentation appearance and organization.


docs/source/index.rst
```diff
@@ -57,25 +34,38 @@ Citing this work
    journal={bioRxiv},
```
maikia, line 34: in the docs it appears in bibtex format. Is that what we want?

maikia (approved):
LGTM, thanks

docs/source/index.rst
BorisMuzellec: Yes, I meant this as a bibtex entry that people could directly copy-paste :). Alternatively we could provide a download link for the bibtex entry, or direct to another page

PR #61, opened 2023-01-23
Fix add missing `joblib_verbosity` argument in `fit_genewise_dispersions`

BorisMuzellec:
This PR adds the `joblib_verbosity` attribute of `DeseqDataSet`, which allows to control the level of verbosity of `Parallel` worker pools, as an argument in the main loop of `fit_genewise_dispersions`, from which it was missing.


maikia (approved):
LGTM

PR #62, opened 2023-01-24
BUG fix DeseqStats test of `dds`'s `replaced` attribute

BorisMuzellec:
At initialization, a DeseqStats object should check that that the DeseqDataSet it is built from has a `replaced` attribute, *if* it has `refit_cooks=True`.

Up to now, it was checking this independently of `dds.refit_cooks`. This small PR corrects this bug.


maikia (commented):
would it be feasible to add a test to it?

BorisMuzellec:
Checking that the workflow `dds.deseq2()` followed with `stats.summary()` when `refit_cooks=False` runs error-free should allow to test that this does no longer happen.

maikia (approved):
Thanks @BorisMuzellec LGTM

PR #64, opened 2023-01-26
DOC Community

maikia:
Some files which might be important for the community as per list here: https://github.com/owkin/PyDESeq2/community

I will add issues templates as separate PRs as it's easier to do from github directly


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,7 @@
+Code of Conduct
+---------------
+
+We want everyone in the community of users and contributors to PyDeseq2.
```
BorisMuzellec, line 4: Seems like there is a missing verb here

BorisMuzellec (changes requested):
LGTM (up to the missing verb / words in the code of conduct)

BorisMuzellec (approved):
Thanks!

PR #72, opened 2023-02-09
Moving docs' requirements into the actual docs' requirements file

mandreux-owkin:
#### Reference Issue or PRs

This PR is not linked to an existing issue.

This PR moves the specific requirements of the documentation (`matplotlib` and `gitpython`) which are currently in the `setup.py` file to the `docs/requirements.txt` file, which is more appropriate as they are not used elsewhere.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -10,3 +10,5 @@ docutils==0.16
 sphinx_click==3.1.0
 sphinx_gallery==0.11.1
 sphinxcontrib-bibtex==2.5.0
+gitpython>=3.1.27
+matplotlib>=3.6.2  # not sure why sphinx_gallery does not work without it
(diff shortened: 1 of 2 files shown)


maikia (changes requested):
@mandreux-owkin thanks for your contribution!
the cleaning up is a great idea, and it's true that gitpython and matplotlib should not be in setup.py if they are only used for docs. 
but here it might be a bit more complex:
Sphinx is generating the docs, but we also use binder to run the example notebooks. (refer to .mybinder directory) There, in postbuild we are actually uninstalling all the docs/requirements to have a lighter build. I am afraid with this change the binder won't work. Feel free to suggest alternative solution.

mandreux-owkin:
@maikia thanks for your careful review. I checked the notebooks code and I am certain that `gitpython` is not used. We don't use any plot in the jupyter but I added `matplotlib` again to avoid breaking binder

maikia (approved):
LGTM, thanks @mandreux-owkin

PR #79, opened 2023-02-15
ENH implement a getter for normalized counts

### LINKED ISSUE #77: How to get normalized expression values
First, thanks for the amazing and much needed work.
In DeSeq2 its possible to output vst normalized expression values as output. Is there such possibility with pyDeseq2?

BorisMuzellec:
#### Reference Issue or PRs

Fixes #77.

This PR implements a getter for `DeseqDataSet`'s normalized counts.

Thanks to the `@property` decorator, normalized counts can be obtained as if they were a `DeseqDataSet` attribute (`dds.normed_counts`), without having to store them in memory. Instead, they are computed on-the-fly.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -238,6 +238,21 @@ class DeseqDataSet:
             # for genes that had outliers replaced
             self.refit()
 
+    @property
+    def normed_counts(self):
+        """
+        Return median-of-ratios normalized counts.
+
+        Called like an attribute (``dds.normed_counts``) thanks to the
+        ``@property decorator``.
+        """
+
+        # Check that size factors are available. If not, compute them.
+        if not hasattr(self, "size_factors"):
+            self.fit_size_factors()
+
+        return self.counts.div(self.size_factors, 0)
+
     def fit_size_factors(self):
         """Fit sample-wise deseq2 normalization (size) factors.
 
@@ -367,7 +382,7 @@ class DeseqDataSet:
 
         print("Fitting dispersion trend curve...")
         start = time.time()
-        self._normed_means = self.counts.div(self.size_factors, 0).mean(0)
+        self._normed_means = self.normed_counts.mean(0)
 
         # Exclude all-zero counts
         targets = self.genewise_dispersions.loc[self.non_zero_genes].copy()


pydeseq2/dds.py
```diff
@@ -238,6 +238,21 @@ def deseq2(self):
             # for genes that had outliers replaced
             self.refit()
 
+    @property
+    def normed_counts(self):
+        """
+        Return median-of-ratios normalized counts.
```
mandreux-owkin, line 244: What do you think about adding a math formula to explain what this median of ratios exactly is, with respect to the original counts?

pydeseq2/dds.py
```diff
@@ -238,6 +238,21 @@ def deseq2(self):
...
+
+        Called like an attribute (``dds.normed_counts``) thanks to the
+        ``@property`` decorator.
+        """
+
+        # Check that size factors are available. If not, compute them.
+        if not hasattr(self, "size_factors"):
+            self.fit_size_factors()
+
+        return self.counts.div(self.size_factors, 0)
```
mandreux-owkin, line 254:
```suggestion
        return self.counts.div(self.size_factors, axis=0)
```

As `axis` is an optional argument of the [`div` method](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.div.html)?

pydeseq2/dds.py
```diff
@@ -238,6 +238,21 @@ def deseq2(self):
             # for genes that had outliers replaced
             self.refit()
 
+    @property
+    def normed_counts(self):
```
mandreux-owkin, line 242: To increase our coverage, let's add a unit test for this method?

mandreux-owkin (commented):
Thanks @BorisMuzellec ! I have tiny remarks

pydeseq2/dds.py
BorisMuzellec: Why not, but I think it would be more relevant to write the formula in `preprocessing.py`, where the code to compute normalisation factors can be found (particularly so as a readable formula won't be a one-liner).

pydeseq2/dds.py
BorisMuzellec: Do you have an idea of what should be tested?

pydeseq2/dds.py
mandreux-owkin: OK we can do this, in this case can you add a reference to relevant method in this docstring? In a "See Also" style, so as to point new users to the correct explanation?

pydeseq2/dds.py
mandreux-owkin: You could test that the value is indeed computed on the fly and returns the correct result (e.g. with a numpy conversion), like

```python
def test_normed_counts():
    dds =DeseqDataset(...)
   counts_np = dds.counts.to_numpy()
    for _ in range(2):
        dds.size_factors = np.random.rand(...)
       expected_normed_counts = counts_np / dds.size_factors
       assert np.allclose(dds.normed_counts.to_numpy(), expected_normed_counts)
```

BorisMuzellec:
Actually, I with the AnnData integration I think that the most natural way to access normalized counts is by setting a new layer, although this induces a memory overhead

pydeseq2/dds.py
mandreux-owkin: OK then this is completely stale!

mandreux-owkin (approved):
LGTM! Much simpler now

PR #68, opened 2023-02-02
ENH Support type hinting using `mypy`

illumination-k:
#### Reference Issue or PRs

Please see #39

#### What does your PR implement? Be specific.

This PR support type hinting by `mypy` and fixed some docstring which incompatible with `mypy` type checking.

PR #80, opened 2023-02-15
CI update workflow

maikia:
update workflow to run pre-commit to avoid redundant code (in workflow and in precommit)

#### Reference Issue or PRs

when adding mypy to pre-commit, CI should automatically run it as well #68 

It removes the checking of the linting from the workflow and instead it runs pre-commit which has those checks included


BorisMuzellec (approved):
LGTM, thanks @maikia!

PR #81, opened 2023-02-15
CI add more versions of python to the github workflow

maikia:
Python 3.8 is supported until October 2024. To ensure the long term support we need to start testing for newer versions of Python as well

#### Reference Issue or PRs

NA

CI check for python 3.9 and 3.10 and 3.11


BorisMuzellec (approved):
LGTM, thanks for thinking ahead!

PR #83, opened 2023-02-20
DOC fix broken RTD example gallery

BorisMuzellec:

This PR fixes the example gallery which causes the docs not to compile on readthedocs. This is due to a data importing issue after merging #68 (pandas won't load a csv from an url provided in the form of a `Path`).

A unit test was added to `test_utils.py` to check that `load_example_data` correctly retrieves data from github when the `datasets` directory is not accessible (as when building the docs on RTD).

### DIFF
--- a/pydeseq2/utils.py
+++ b/pydeseq2/utils.py
@@ -73,16 +73,16 @@ def load_example_data(
     if dataset == "synthetic":
         path_to_data = datasets_path / "synthetic"
         if Path(path_to_data).is_dir():
-            path_to_data_counts = path_to_data / "test_counts.csv"
-            path_to_data_clinical = path_to_data / "test_clinical.csv"
+            path_to_data_counts = str(path_to_data / "test_counts.csv")
+            path_to_data_clinical = str(path_to_data / "test_clinical.csv")
         else:
             # if the path does not exist (as is the case in RDT) load it from github
             url_to_data = (
                 "https://raw.githubusercontent.com/owkin/"
                 "PyDESeq2/main/datasets/synthetic/"
             )
-            path_to_data_counts = Path(url_to_data + "/test_counts.csv")
-            path_to_data_clinical = Path(url_to_data + "/test_clinical.csv")
+            path_to_data_counts = url_to_data + "/test_counts.csv"
+            path_to_data_clinical = url_to_data + "/test_clinical.csv"
 
         if modality == "raw_counts":
             df = pd.read_csv(
--- a/pydeseq2/utils.py
+++ b/pydeseq2/utils.py
@@ -73,6 +73,8 @@ def load_example_data(
     if dataset == "synthetic":
         path_to_data = datasets_path / "synthetic"
         if Path(path_to_data).is_dir():
+            # Cast the Paths to strings to have coherent types wrt to the url case (that
+            # does not handle Paths), else mypy throws an error.
             path_to_data_counts = str(path_to_data / "test_counts.csv")
             path_to_data_clinical = str(path_to_data / "test_clinical.csv")
         else:
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -35,8 +35,9 @@ def test_nb_nll_moments(mu, alpha):
 
 # Test data loading from outside the package (e.g. on RTF)
 @pytest.mark.parametrize("modality", ["raw_counts", "clinical"])
+@pytest.mark.parametrize("mocked_dir_flag", [True, False])
 @mock.patch("pathlib.Path.is_dir")
-def test_rtd_example_data_loading(mocked_function, modality):
+def test_rtd_example_data_loading(mocked_function, modality, mocked_dir_flag):
     """
     Test that load_example_data still works when run from a place where the ``datasets``
     directory is not accessible, as is when the documentation is built on readthedocs.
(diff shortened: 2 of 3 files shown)


tests/test_utils.py
```diff
@@ -27,3 +31,29 @@ def test_nb_nll_moments(mu, alpha):
     assert np.abs(diff) < 0.2 * deviation
     error_var = np.abs(sample.var() - var_th) / var_th
     assert error_var < 1 / np.sqrt(n_montecarlo)
+
+
+# Test data loading from outside the package (e.g. on RTF)
```
maikia, line 36:
```suggestion
# Test data loading from outside the package (e.g. on RTF)
@pytest.mark.parametrize("return_value", [True, False])
```

why not to directly test for both values? and set return_value to the that parameter

tests/test_utils.py
```diff
@@ -27,3 +31,29 @@ def test_nb_nll_moments(mu, alpha):
...
+    Test that load_example_data still works when run from a place where the ``datasets``
+    directory is not accessible, as is when the documentation is built on readthedocs.
+    """
+
+    # Mock the output of is_dir() as False to emulate not having access to the
+    # ``datasets`` directory
+    pathlib.Path.is_dir.return_value = False
+
+    # Try loading data. Ignore flake8 "variable is assigned to but never used" error.
+    counts_df = load_example_data(  # noqa: F841
```
maikia, line 49: instead of ignoring (noqa) don't assign any value (unless you want to check it for correctness in type, shape or anything else?

tests/test_utils.py
```diff
@@ -27,3 +31,29 @@ def test_nb_nll_moments(mu, alpha):
...
+        modality="raw_counts",
+        dataset="synthetic",
+        debug=False,
+    )
+
+    clinical_df = load_example_data(  # noqa: F841
```
maikia, line 55: same here

pydeseq2/utils.py
```diff
@@ -73,16 +73,16 @@ def load_example_data(
     if dataset == "synthetic":
         path_to_data = datasets_path / "synthetic"
         if Path(path_to_data).is_dir():
-            path_to_data_counts = path_to_data / "test_counts.csv"
-            path_to_data_clinical = path_to_data / "test_clinical.csv"
+            path_to_data_counts = str(path_to_data / "test_counts.csv")
```
maikia, line 76: I tested if it passes for the is_dir True when you remove changing to string and it passes (path_to_data_counts is <class 'pathlib.PosixPath'> and that's ok for numpy. ie no need to set to str in this case. The numpy fails if url is a Path (which makes sense) but if it's a path within the directories it's fine

tests/test_utils.py
```diff
@@ -27,3 +31,29 @@ def test_nb_nll_moments(mu, alpha):
...
+    directory is not accessible, as is when the documentation is built on readthedocs.
+    """
+
+    # Mock the output of is_dir() as False to emulate not having access to the
+    # ``datasets`` directory
+    pathlib.Path.is_dir.return_value = False
+
+    # Try loading data. Ignore flake8 "variable is assigned to but never used" error.
+    counts_df = load_example_data(  # noqa: F841
+        modality="raw_counts",
```
maikia, line 50: why not make 'modality' a parametrize of pytest as well?

maikia (changes requested):

pydeseq2/utils.py
BorisMuzellec: Indeed, but `mypy` is not happy when `path_to_data_counts` can either be a `Path` or a `str`, so I cast it to `str` which is the common denominator

tests/test_utils.py
BorisMuzellec: indeed!

tests/test_utils.py
BorisMuzellec: Can this work in conjunction with `mocking`?

pydeseq2/utils.py
maikia: ok, in this case maybe just add a comment on why it is done like this

tests/test_utils.py
maikia: I don't see why not. but test it ;-)

BorisMuzellec: Done! (in [fe6d79f](https://github.com/owkin/PyDESeq2/pull/83/commits/fe6d79fc2a89ee34bd8524cdc87fc11cd610e25d))

pydeseq2/utils.py
BorisMuzellec: Done

tests/test_utils.py
```diff
@@ -27,3 +31,24 @@ def test_nb_nll_moments(mu, alpha):
...
+@mock.patch("pathlib.Path.is_dir")
+def test_rtd_example_data_loading(mocked_function, modality):
+    """
+    Test that load_example_data still works when run from a place where the ``datasets``
+    directory is not accessible, as is when the documentation is built on readthedocs.
+    """
+
+    # Mock the output of is_dir() as False to emulate not having access to the
+    # ``datasets`` directory
+    pathlib.Path.is_dir.return_value = [False, True]
```
maikia, line 47: this should be in parametrize (otherwise the value is [False, True]

maikia (changes requested):

tests/test_utils.py
BorisMuzellec: Indeed... would this work [fa707b9](https://github.com/owkin/PyDESeq2/pull/83/commits/fa707b9da68b88c5326063d40aba2fce96fc2c18)?

maikia (approved):
LGTM, thanks @BorisMuzellec

PR #90, opened 2023-02-27
BUG fix replacement counts when there are un-replaceable samples

BorisMuzellec:

This PR fixes a bug (incompatible boolean mask and array shapes, encountered when testing on TCGA-PAAD) which appeared when trying to refit cooks outliers for datasets in which some samples are un-replaceable due to not having enough replicates.

It also implements a test to ensure that the code runs error-free in that setting.


mandreux-owkin (approved):
LGTM!

PR #105, opened 2023-03-28
BUG dispersion prior variance

mikelove:
In the dispersion prior variance step, the expected variance of the log dispersion involves a function of (m - p), number of samples minus number of columns of model matrix. Compare to:

https://github.com/mikelove/DESeq2/blob/devel/R/core.R#L1182

I'm trying to align some of the parameters a bit.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -513,7 +513,6 @@ class DeseqDataSet(ad.AnnData):
             self.fit_dispersion_trend()
 
         # Exclude genes with all zeroes
-        # num_genes = len(self.non_zero_genes)
         num_samples = self.n_obs
         num_vars = self.obsm["design_matrix"].shape[-1]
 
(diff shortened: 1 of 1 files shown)


pydeseq2/dds.py
```diff
@@ -513,7 +513,8 @@ def fit_dispersion_prior(self) -> None:
             self.fit_dispersion_trend()
 
         # Exclude genes with all zeroes
-        num_genes = len(self.non_zero_genes)
+        # num_genes = len(self.non_zero_genes)
```
BorisMuzellec, line 516:
```suggestion
```

BorisMuzellec (approved):

mandreux-owkin:
Hi @mikelove 

> how can I do local development?

Instead of `pip install`, just do `pip install -e .` in repository you cloned, this puts the package in editable mode.

PR #109, opened 2023-04-04
ENH  set reference level in `DeseqDataSet`

BorisMuzellec:

This PR allows users to set the desired reference (control) level when creating a `DeseqDataSet`, instead of only the tested level as previously.

This ensures that users have control over the LFC coefficients created by a `DeseqDataSet`, so that they can ensure that the relevant coefficients exist when applying `lfc_shrink`, which wasn't the case when only the tested level could be set.

### DIFF
--- a/tests/test_pydeseq2.py
+++ b/tests/test_pydeseq2.py
@@ -471,4 +471,12 @@ def test_ref_level():
         ref_level=["group", "Y"],
     )
 
+    # Check that the column exists
     assert "group_X_vs_Y" in dds.obsm["design_matrix"].columns
+    # Check that its content is correct
+    assert (
+        dds.obsm["design_matrix"]["group_X_vs_Y"]
+        == clinical_df["group"].apply(
+            lambda x: 1 if x == "X" else 0 if x == "Y" else np.NaN
+        )
+    ).all()


tests/test_pydeseq2.py
```diff
@@ -446,3 +446,29 @@ def test_vst(tol=0.02):
...
+    )
+
+    dds = DeseqDataSet(
+        counts=counts_df,
+        clinical=clinical_df,
+        design_factors=["group", "condition"],
+        ref_level=["group", "Y"],
+    )
+
+    assert "group_X_vs_Y" in dds.obsm["design_matrix"].columns
```
maikia, line 474: do you think it might be worth for checking if it is correct? I still accept and leave the decision to you

maikia (approved):
LGTM, thanks

tests/test_pydeseq2.py
BorisMuzellec: You mean checking whether the column contains correct data? It's true that this only checks that it exists.

maikia: Yes, that's what I was thinking about. Would it be difficult to add?

PR #110, opened 2023-04-05
BUG fix slogdet bug

BorisMuzellec:

This PR fixes bugs when using `pydeseq2` with `pandas` 2.0.

##### First bug

```
counts_df = load_example_data(
    modality="raw_counts",
    dataset="synthetic",
    debug=False,
)

clinical_df = load_example_data(
    modality="clinical",
    dataset="synthetic",
    debug=False,
)


dds = DeseqDataSet(
    counts=counts_df,
    clinical=clinical_df,
    refit_cooks=True,
    design_factors="condition",
)

dds.deseq2()
```

would output

```
Fitting size factors...
... done in 0.00 seconds.

Fitting dispersions...

UFuncTypeError: Cannot cast ufunc 'slogdet' input from dtype('O') to dtype('float64') with casting rule 'same_kind'
```

##### Second bug

After fixing the first bug, `DeseqDataSet.refit()` would throw `anndata.ImplicitModificationWarning`s, making the tests fail. 

#### Fix

The first bug is caused by `pydeseq2.utils.build_design_matrix` outputting boolean columns with `pandas` API changes. This PR fixes it by casting the design matrix to int, which is compatible with older versions of `pandas`.

The second bug arrises when copying anndata objects with empty indexes. This PR fixes it by checking that indexes are non-empty prior to copying.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -867,8 +867,6 @@ class DeseqDataSet(ad.AnnData):
         # Only replace if genes are not all zeroes after outlier replacement
         to_replace[to_replace] = ~new_all_zeroes
 
-        print(to_replace)
-
         self.varm["_normed_means"][to_replace] = sub_dds.varm["_normed_means"]
         self.varm["LFC"][to_replace] = sub_dds.varm["LFC"]
         self.varm["dispersions"][to_replace] = sub_dds.varm["dispersions"]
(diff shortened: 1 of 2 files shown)


pydeseq2/dds.py
```diff
@@ -866,6 +867,8 @@ def _refit_without_outliers(
         # Only replace if genes are not all zeroes after outlier replacement
         to_replace[to_replace] = ~new_all_zeroes
 
+        print(to_replace)
```
maikia, line 870: remove?

maikia (approved):
LGTM, thanks

pydeseq2/dds.py
BorisMuzellec: Indeed, you caught me red-handed doing print debugging :)

PR #123, opened 2023-04-28
Implement fitType="mean" for vst

### LINKED ISSUE #116: Implement fitType="mean" for vst
DESeq2 has a fit type "mean" for the vst which is used as the fallback when the parametric fit fails. It would be nice to have an equivalent in the python implementation.

Here's my attempt based on https://github.com/mikelove/DESeq2/blob/devel/R/vst.R#L188 and https://github.com/mikelove/DESeq2/blob/devel/R/core.R#L892

```
        gene_dispersions = self.varm["genewise_dispersions"]
        use_for_mean = gene_dispersions > 10*self.min_disp
        mean_disp = scipy.stats.trim_mean(gene_dispersions[use_for_mean], proportiontocut=0.001)
        self.layers["vst_counts"] = ( 2 * np.arcsinh(np.sqrt(mean_disp * self.layers["normed_counts"])) - np.log(mean_disp) - np.log(4) ) / np.log(2)
```

scottgigante-immunai:
Closes #116.

DESeq2 has a fit type "mean" for the vst which is used as the fallback when the parametric fit fails. It would be nice to have an equivalent in the python implementation.

Here's my attempt based on https://github.com/mikelove/DESeq2/blob/devel/R/vst.R#L188 and https://github.com/mikelove/DESeq2/blob/devel/R/core.R#L892 .

No guarantees that it's exactly the same -- I'm not really familiar enough with the algorithm to dive much deeper.


scottgigante-immunai:
@BorisMuzellec this is failing because IPython no longer supports Python3.8, unrelated to my PR. Leaving it to you to write tests to compare performance here to the R version.

BorisMuzellec:
Thanks for this PR @scottgigante-immunai! I'll try to find some time this week to fix the IPython issue and write some R / Python tests.

BorisMuzellec (approved):
Just implemented tests to compare with DESeq2, seems all good! Thanks @scottgigante-immunai! I'm merging this PR

scottgigante-immunai:
Amazing! Thanks @BorisMuzellec

PR #131, opened 2023-05-10
BUG wrong `lfc_shrink` sign display

### LINKED ISSUE #121: [BUG] `pydeseq2.ds.DeseqStats.lfc_shrink` gives wrong log2FoldChange signs
**Describe the bug**
When performing LFC shrinkage, the new version seems to flip the reference it uses. I've observed that the obtained `log2FoldChange` have opposite signs as the original contrast, which creates an inconsistency with the sign of `stat`.

**To Reproduce**
You can reproduce this bug with `pydeseq2 >= 0.3.1`. Here's a reproducible example:
```python
import pandas as pd
import numpy as np
from anndata import AnnData
from pydeseq2.dds import DeseqDataSet
from pydeseq2.ds import DeseqStats

# Build toy data, note that G1 should be up in healthy, G2 does not change and G3 should be up in disease
var = pd.DataFrame(index=['G1', 'G2', 'G3'])
X = pd.DataFrame(
    [[16., 7., 12.],
     [22., 6., 1.],
     [10., 8., 30.],
     [3., 6., 20.],
     [1., 5., 17.]], columns=var.index
)
obs = pd.DataFrame(
    [['healthy'],
     ['healthy'],
     ['disease'],
     ['disease'],
     ['disease']], columns=['condition'], index=['S1', 'S2', 'S3', 'S4', 'S5']
    )
adata = AnnData(X, obs=obs, var=var, dtype=np.float32)

# Build DESeq2 object
dds = DeseqDataSet(
    adata=adata,
    design_factors='condition',
    refit_cooks=True,
    n_cpus=8,
)

# Compute LFCs
dds.deseq2()

# Extract contrast between disease vs healthy
stat_res = DeseqStats(dds, contrast=["condition", 'disease', 'healthy'], n_cpus=8)

# Compute Wald test
stat_res.summary()

# Store results before shrink
before_shrink = stat_res.results_df

# Assert that logFCs should have the same sign than stat
assert np.all(np.sign(before_shrink['log2FoldChange']) == np.sign(before_shrink['stat']))

# Shrink LFCs
stat_res.lfc_shrink(coeff='condition_healthy_vs_disease')

after_shrink = stat_res.results_df

# Assert that logFCs should have the same sign than stat
assert np.all(np.sign(after_shrink['log2FoldChange']) == np.sign(after_shrink['stat']))
```

**Expected behavior**
After LFC shrinkage, `log2FoldChange` should have the same sign than `stat`.

**Screenshots**
Before shrinkage:
image

After shrinkage:
image

**Desktop (please complete the following information):**
 - OS: linux

**Additional context**
May I ask why now the method `pydeseq2.ds.DeseqStats.lfc_shrink` requires `coeff`? Before (between version `0.3.0` and `0.3.1`) it could be run based on the contrast specified already in the `DeseqStats` object. Now you need to manually check `stat_res.LFC` for the name which feels weird.

I'm aware that if you provide `ref_level=['condition', 'healthy']` to the previous `DeseqDataSet` it will work but then there are two times where the contrast needs to be specified and if there are discrepancies, like here, errors will rise. IMO it makes more sense to specify the contrast at the `DeseqStats` level instead of the `DeseqStats` one.

BorisMuzellec:
#### Reference Issue or PRs
Fixes #121 

This PR addresses #121 (wrong display of shrunk factors). More precisely:
- the message output by `lfc_shrink` was fixed to match the coefficient that was actually shrunk
- when the coefficient is not available, the error message now explains how to set `ref_level` in `DeseqDataSet`
- it is now possible to set `coeff=None`
- `coeff=None` is tested in `test_edge_cases.py`

### DIFF
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -386,6 +386,8 @@ class DeseqStats:
             # Get the corrresponding factor, tested and reference levels of the shrunk
             # coefficient
             split_coeff = coeff.split("_")
+            # coeffs are of the form "factor_A_vs_B", hence "factor" is split_coeff[0],
+            # "A" is split_coeff[1] and "B" split_coeff[3]
             print(
                 f"Shrunk Log2 fold change & Wald test p-value: "
                 f"{split_coeff[0]} {split_coeff[1]} vs {split_coeff[3]}"
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -275,7 +275,8 @@ class DeseqStats:
         Parameters
         ----------
         coeff : str or None
-            The LFC coefficient to shrink.
+            The LFC coefficient to shrink. If set to ``None``, the method will try to
+            shrink the coefficient corresponding to the ``contrast`` attribute.
             If the desired coefficient is not available, it may be set from the
             :class:`pydeseq2.dds.DeseqDataSet` argument ``ref_level``.
             (default: ``None``).
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -267,25 +267,43 @@ class DeseqStats:
             self.statistics.loc[self.dds.new_all_zeroes_genes] = 0.0
             self.p_values.loc[self.dds.new_all_zeroes_genes] = 1.0
 
-    def lfc_shrink(self, coeff: str) -> None:
+    def lfc_shrink(self, coeff: Optional[str] = None) -> None:
         """LFC shrinkage with an apeGLM prior :cite:p:`DeseqStats-zhu2019heavy`.
 
         Shrinks LFCs using a heavy-tailed Cauchy prior, leaving p-values unchanged.
 
         Parameters
         ----------
-        coeff : str
+        coeff : str or None
             The LFC coefficient to shrink.
             If the desired coefficient is not available, it may be set from the
             :class:`pydeseq2.dds.DeseqDataSet` argument ``ref_level``.
+            (default: ``None``).
         """
 
-        if coeff not in self.LFC.columns:
+        contrast_level = f"{self.contrast[0]}_{self.contrast[1]}_vs_{self.contrast[2]}"
+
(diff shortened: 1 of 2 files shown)


pydeseq2/ds.py
```diff
@@ -356,9 +383,12 @@ def lfc_shrink(self, coeff: str) -> None:
             self.results_df["log2FoldChange"] = self.LFC.iloc[:, coeff_idx] / np.log(2)
             self.results_df["lfcSE"] = self.SE / np.log(2)
 
+            # Get the corrresponding factor, tested and reference levels of the shrunk
+            # coefficient
+            split_coeff = coeff.split("_")
             print(
                 f"Shrunk Log2 fold change & Wald test p-value: "
-                f"{self.contrast[0]} {self.contrast[1]} vs {self.contrast[2]}"
+                f"{split_coeff[0]} {split_coeff[1]} vs {split_coeff[3]}"
```
maikia, line 391: I got a bit lost, I was expecting `split_coeff[2]` instead of 3. can you write a short comment so it's easier to follow on what's in the `coeff` (or `split_coeff`) when reading the code?

tests/test_edge_cases.py
```diff
@@ -167,6 +167,8 @@ def test_lfc_shrinkage_coeff():
 
     res = DeseqStats(dds)
     res.summary()
+    res.lfc_shrink(coeff=None)
```
maikia, line 170: would it be worth adding a test for other coeff values?

BorisMuzellec: I've updated the test to check that the same result is obtained with the implicit coeff `condition_B_vs_A` (the case where an incorrect coeff is provided is already covered)

BorisMuzellec:
Thanks for your review @maikia! Could you have a look at the updates I made and see if they answer the points you raised?

maikia (approved):
LGTM, thanks @BorisMuzellec

PR #135, opened 2023-06-02
cron scheduler: CI runs every week day at 3:30 UTC

mandreux-owkin:

This PR implements a nightly CI run on every week day. Currently, the CI runs only when pull requests are submitted, which is infrequent. This prevents us from catching bugs due to version bumps early. These nightly tests will help developers catch them as soon as they arise.


BorisMuzellec (approved):
LGTM

Issue #130, opened 2023-05-05
[BUG] Issue with contrast if underscore is in sample names


mousepixels:
Replaced the '_' with '' in R_bleo and it fixed the issue. Interestingly, if I contrast 'R_bleo" vs "No_bleo" I do not get an error.

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~/miniconda3/envs/sc/lib/python3.9/site-packages/pandas/core/indexes/base.py:3652, in Index.get_loc(self, key)
   3651 try:
-> 3652     return self._engine.get_loc(casted_key)
   3653 except KeyError as err:

File ~/miniconda3/envs/sc/lib/python3.9/site-packages/pandas/_libs/index.pyx:147, in pandas._libs.index.IndexEngine.get_loc()

File ~/miniconda3/envs/sc/lib/python3.9/site-packages/pandas/_libs/index.pyx:176, in pandas._libs.index.IndexEngine.get_loc()

File pandas/_libs/hashtable_class_helper.pxi:7080, in pandas._libs.hashtable.PyObjectHashTable.get_item()

File pandas/_libs/hashtable_class_helper.pxi:7088, in pandas._libs.hashtable.PyObjectHashTable.get_item()

KeyError: 'Sample_R_bleo_vs_bleo'

The above exception was the direct cause of the following exception:

KeyError                                  Traceback (most recent call last)
Cell In[306], line 1
----> 1 stat_res = DeseqStats(dds, n_cpus=8, contrast=('Sample', 'R_bleo', 'pCD47'))

File ~/miniconda3/envs/sc/lib/python3.9/site-packages/pydeseq2/ds.py:150, in DeseqStats.__init__(self, dds, contrast, alpha, cooks_filter, independent_filter, n_cpus, prior_LFC_var, batch_size, joblib_verbosity)
    147 self.LFC = self.dds.varm["LFC"].copy()
    149 # Build a contrast vector corresponding to the variable and levels of interest
--> 150 self._build_contrast_vector()
    152 # Set a flag to indicate that LFCs are unshrunk
    153 self.shrunk_LFCs = False

File ~/miniconda3/envs/sc/lib/python3.9/site-packages/pydeseq2/ds.py:595, in DeseqStats._build_contrast_vector(self)
    589 else:
    590     # Need to change reference
    591     # Get any column corresponding to the desired factor and extract old ref
    592     old_ref = next(
[...]

PR #141, opened 2023-06-09
fix empty `old_ref` bug caused by split `_`

hermanzhaozzzz:
#### unittest
no need

If underscores contain in the sample name, the `next()` step will fail.
I fixed the empty `old_ref` bug caused by split `_`

Replicating this BUG is as easy as adding multiple underscores to the sample

Run info as follow:

Xnip2023-06-09_15-35-20

as you can see, the `old_ref` var is empty cased a empty return by `next`

Xnip2023-06-09_15-35-29
Xnip2023-06-09_15-40-14

If there are other similar problems, I hope the author will consider modifying them, after all, for a coder writing python, it is very common to use underscores in sample names


BorisMuzellec:
Hi @hermanzhaozzzz, thank you for opening a PR.

I think that your suggestion does the trick in this particular method, but there could be other places where underscores cause bugs. I'll update the test to check whether this is the case.

BorisMuzellec:
Fixes #130

BorisMuzellec:
Fixes #160

pydeseq2/dds.py
```diff
@@ -220,6 +220,39 @@ def __init__(
...
+                },
+                inplace=True,
+            )
+
+            self.design_factors = new_factors
+
+            # Also check continuous factors
+            if self.continuous_factors is not None:
+                self.continuous_factors = [
+                    factor.replace("_", "-") for factor in self.continuous_factors
```
maikia, line 248: It seems like you are doing the same thing for multiple variables. Perhaps you could write a helper function where you just pass those variables, make the switch and return it back instead of repeating the same each time?

maikia (approved):
LGTM, thanks @BorisMuzellec

pydeseq2/dds.py
BorisMuzellec: Indeed! I've added a `replace_underscores` function in `utils.py`.

BorisMuzellec:
Thanks again @hermanzhaozzzz for opening this PR, and thanks @maikia for the review!

hermanzhaozzzz:
Thanks for your work!

——————————————————————
ZHAO Hua-nan

Ph. D. candidate
School of Life Sciences, Tsinghua University

Ph. D. Joint training
School of Life Sciences, Peking University



在 2023年8月7日 +0800 20:58，Boris Muzellec ***@***.***>，写道：

Issue #102, opened 2023-03-22
ENH quiet mode


BorisMuzellec:
**Is your feature request related to a problem? Please describe.**
The `DeseqDataSet` and `DeseqStats` classes print a message at each step of the pipeline (e.g. fitting size factors, or dispersions), including when those steps are nested within another task (e.g., computing VST #99), which can lead to many useless statements being printed.

**Describe the solution you'd like**
Having a verbosity parameter that would allow controlling the desired level of verbosity of `fit_size_factors`, `fit_genewise_dispersions`, `fit_dispersion_trend`, and so on.  

**Describe alternatives you've considered**
Deactivating print statements from nested calls (e.g. VST).

**Additional context**
Currently, computing VST (#99) prints the following:
![Capture d’écran 2023-03-22 à 09 59 40](https://user-images.githubusercontent.com/11092154/226852219-7508d1a8-7040-44de-8b17-9b9f85a0c308.png)

dbdimitrov:
Hi,

Indeed, verbosity would be fantastic to have.
It might be also worthwhile to use the same quiet flag for the `display` calls.

In the meantime, `%%capture` seems to the the trick for me.
https://stackoverflow.com/questions/23692950/how-do-you-suppress-output-in-jupyter-running-ipython

BorisMuzellec:
Fixed in #146

PR #146, opened 2023-07-01
Introduce quiet attribute to `DeseqDataSet` that is checked before printing status updates

noamteyssier:
#### Reference Issue or PRs

This is in reference to issue #102, i.e. introducing a quiet parameter to suppress status updates.

1. This introduces a `self.quiet` bool attribute to the `DeseqDataSet` class
2. Functions that `print` first check if the `self.quiet` attribute is set and only print if so
3. Print statements are routed to `sys.stderr` as opposed to stdout (this wasn't in the issue but I find this to be more standard for status updates like this)

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -180,7 +180,6 @@ class DeseqDataSet(ad.AnnData):
         joblib_verbosity: int = 0,
         quiet: bool = False,
     ) -> None:
-
         # Initialize the AnnData part
         if adata is not None:
             if counts is not None:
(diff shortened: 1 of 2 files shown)


BorisMuzellec:
Hi @noamteyssier, thanks a lot for this PR, looks great!

To make this PR complete, it would be nice if you had the time to extend it to `DeseqStats` (in `dds.py`) as well. Happy to merge after that :)

noamteyssier:
No problem - just extended it to the `DeseqStats` class as well.

BorisMuzellec (approved):
Thanks, LGTM!

PR #153, opened 2023-07-19
DOC Degrees of freedom corner case

BorisMuzellec:

As pointed out by @mikelove, in the corner case where the design matrix has fewer than 3 degrees of freedom, the distribution of log dispersions is poorly estimated by the MAD, as in `fit_dispersion_prior` (cf the [DESeq2](https://github.com/mikelove/DESeq2/blob/24099effd42f31b96bf7c867ea03d7300a6368bc/R/core.R#L631) implementation of `estimateDispersionsPriorVar`).

This is a relatively minor issue that should be addressed in the future. Meanwhile, this PR:

- updates the docstring of `fit_dispersion_prior` to reflect this issue,
- raises a warning when this corner case arises.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -557,6 +557,9 @@ class DeseqDataSet(ad.AnnData):
         """Fit dispersion variance priors and standard deviation of log-residuals.
 
         The computation is based on genes whose dispersions are above 100 * min_disp.
+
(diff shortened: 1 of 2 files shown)


maikia (approved):
Thanks @BorisMuzellec 
LGTM.

PR #163, opened 2023-08-01
ENH continuous covariates

### LINKED ISSUE #100: Multi-level factors and continuous covariates
Hi devs,

Amazing work! It's really nice to have such a strong DEA framework finally available in python. Here's a suggestion regarding factors:

**Is your feature request related to a problem? Please describe.**
Currently pydeseq2 is limited to bi-level factors, limiting its usage whenever there are multi-level factors or continuous covariates.

**Describe the solution you'd like**
I know it might be cumbersome to implement, but it would be convenient to add support to multi-level factors and continuous covariates in the `design_factors` argument of DeseqDataSet.

**Describe alternatives you've considered**
For multi-level factors, the current alternative is to generate a single `DeseqDataSet` for each combination of groups, making it tedious to deal with (although it's true that depending on the within-group variability differences it might be needed as explained in the [original docs](https://bioconductor.org/packages/release/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#if-i-have-multiple-groups-should-i-run-all-together-or-split-into-pairs-of-groups)).

For continuous covariates, the current alternative is to cut them into bins and generate factors of them, which results in a multi-level factor with the previously mentioned problems.

BorisMuzellec:
#### Reference Issue or PRs
Fixes #100 

This PR adds support for continuous covariates in PyDESeq2.

More precisely:
- An optional `continuous_factors` argument was added to `DeseqDataSet` and `utils.build_design_matrix`, in which users may provide a list of design factors that should be considered as continuous / numerical;
- The contrast handling of `DeseqStats` was updated to account for continuous factors;
- Tests were added to compare the results of PyDESeq2 to those of DESeq2 in the presence of continuous factors, on synthetic data.


pydeseq2/ds.py
```diff
@@ -633,26 +653,38 @@ def _build_contrast(self, contrast: Optional[List[str]] = None) -> None:
...
-                raise KeyError(
-                    f"The tested level ('{contrast[1]}') should correspond to one of the"
-                    f" levels of '{contrast[0]}'"
-                )
-            if contrast[2] not in self.dds.obs[contrast[0]].values:
-                raise KeyError(
-                    f"The reference level ('{contrast[2]}') should correspond to one of"
-                    f" the levels of '{contrast[0]}'"
-                )
+            if not contrast[1] == contrast[2] == "":
```
maikia, line 656:
```suggestion
            if not (contrast[1] == contrast[2] == ""):
```
Just to be clear on the logical statement

pydeseq2/utils.py
```diff
@@ -192,55 +196,79 @@ def build_design_matrix(
...
+                        if f"{factor}_{level}" not in design_matrix.columns
+                    )
+                else:
+                    # The reference level is given as an argument
+                    ref = ref_level[1]
+                design_matrix.columns = [
+                    f"{col}_vs_{ref}" if col.startswith(factor) else col
+                    for col in design_matrix.columns
+                ]
+
```
maikia, line 255:
```suggestion
```

maikia (approved):
LGTM, thanks @BorisMuzellec

BorisMuzellec:
Thanks for the review @maikia

PR #166, opened 2023-08-14
MAINT reduce the number of arguments in ``_fit_rough_dispersions``

BorisMuzellec:

This PR is a minor refactoring of `DeseqDataSet._fit_rough_dispersions()`.

 Currently, this method takes both `counts` and `size_factors` as arguments, only to compute normalized factors as `counts / size_factors[:, None]`.

We now provide `normed_counts` directly, which are already stored in `self.layers["normed_counts"]` when `_fit_rough_dispersions()` is called in the pydeseq2 pipeline.

Likewise, we provide `normed_counts` instead of `counts` in `_fit_moments_dispersions()` to avoid performing the division again.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -826,12 +826,11 @@ class DeseqDataSet(ad.AnnData):
         """
 
         # Check that size_factors are available. If not, compute them.
-        if "size_factors" not in self.obsm:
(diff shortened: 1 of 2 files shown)


maikia (approved):
thanks @BorisMuzellec overall it LGTM

BorisMuzellec:
Thanks @maikia!

PR #171, opened 2023-09-04
ci: test against pre-release versions of dependencies

### LINKED ISSUE #170: Test against pre-release versions of dependencies
Hey! Really like this project, very happy to see DE being brought to python!

I'm wondering if you could add a CI job testing against release candidates of your dependencies. Here's and example of what that looks like over in the [scverse cookiecutter template](https://github.com/scverse/cookiecutter-scverse):

<details>
<summary> Example </summary>

[link](https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/test.yaml#L28-L30)

```yaml
jobs:
  test:
    ...
    strategy:
      matrix:
        include:
          ...
          - os: ubuntu-latest
            python: "3.10"
            pip-flags: "--pre"
    ...

    steps:
      ...
      - name: Install dependencies
        run: |
          pip install ${{ matrix.pip-flags }} ".[dev,test]"
```

</details>

I saw you're subclassing AnnData, which isn't a thing we've thought a lot about people doing, so was a little worried future updates might cause problems for you. It would be really good to know about these things before a real release of AnnData goes out, so that we can figure out what to do then.

I'd be happy to help with this if you'd like.

--------

We are starting to do a little bit of thinking about `AnnData` subclasses. Would you be interested in being a part of those conversations?

BorisMuzellec:
#### Reference Issue or PRs

Closes #170 

Following @ivirshup's suggestion, this PR adds tests against pre-release versions of pydeseq2's dependencies to the CI.

### DIFF
--- a/.github/workflows/pr_validation.yml
+++ b/.github/workflows/pr_validation.yml
@@ -16,8 +16,16 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
-        python: ["3.8", "3.9", "3.10", "3.11"]
+        include:
+          - os: ubuntu-latest
+              python: "3.8"
+          - os: ubuntu-latest
+            python: "3.9"
+          - os: ubuntu-latest
+            python: "3.10"
+          - os: ubuntu-latest
+            python: "3.10"
+            pip-flags: "--pre"
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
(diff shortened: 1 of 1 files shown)


maikia (approved):
I was not aware of the `pip-flags: "--pre"` but seems to be working well. What is the reasoning behind removing Python 3.11? I would prefer if we don't do it. for a reference the table of [supported python versions](https://devguide.python.org/versions/). 
Nevertheless, I approve and let you make the final decision

BorisMuzellec:
Hi @maikia, thanks for having a look. Nice catch re python 3.11, it was not intentional. Let me add it back.

What do you think of removing support / tests for Python 3.8 though? In particular, [recent releases of numpy are no longer supporting it](https://numpy.org/doc/stable/release/1.25.0-notes.html).

PR #188, opened 2023-10-18
Drop iPython dependency and simple print instead

adamgayoso:
IPython is a heavy dependency (especially for non-essential functionality), and it would be appreciated to replace this with something simpler. Thanks!


BorisMuzellec (approved):
Indeed, it's always a good idea to make packages lighter. LGTM, thanks!

PR #191, opened 2023-10-20
Factor out inference routines

### LINKED ISSUE #187: Factoring out the core implementation with a `DESeqImplementation` class
Thank you for porting deseq to python, this is a great implementation!

I'm interested in factoring out some of the core optimization routines into their own class so that alternative multiprocessing routines can be used in a more straightforward way.

For example, a chunk of code like this https://github.com/owkin/PyDESeq2/blob/8fbcc2faf744518852c90552242a5049f5d9c9c8/pydeseq2/ds.py#L427-L449

could be placed inside a class like

```
class DESeqImplementation:

    def __init__(verbosity, batch_size):
        # ...

    def lfc_shrink_nbinom_glm(design_matrix, counts, ...) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
       # core routine
```

along with the `wald_test`, `irls`, `fit_alpha_mle`, etc.

the implementation class would then be a keyword arg to the `__init__` of `DeseqStats` and `DeseqDataSet` and facilitate multiple possible backends for inference.

adamgayoso:
Fixes #187


adamgayoso:
As an aside, it would be helpful to add https://pre-commit.ci/ as a bot so that PRs can be autofixed

adamgayoso:
Just want to again mention the pre-commit here https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.pre-commit-config.yaml

Ruff is really fast and adoption is rising. It can also fix a lot of the things I'm fixing myself right now :)

adamgayoso:
@BorisMuzellec I'm not sure how strongly you feel about the test_docstrings file now that ruff should be handling everything. It's failing due to no extended summary

BorisMuzellec:
> @BorisMuzellec I'm not sure how strongly you feel about the test_docstrings file now that ruff should be handling everything. It's failing due to no extended summary

`test_docstrings` mostly runs `numpydoc.validate`. Given that ruff runs pydocstyle I think there's no need to double-check docstrings indeed

adamgayoso:
@BorisMuzellec it will probably also help the project to use readthedocs pull request builds https://docs.readthedocs.io/en/stable/guides/pull-requests.html

And then building docs can be previewed per PR and removed from the tests. We had done this in scvi

pydeseq2/inference.py
```diff
@@ -0,0 +1,370 @@
...
+        min_mu : float
+            Lower threshold for fitted means, for numerical stability.
+            (default: ``0.5``).
+
+        Returns
+        -------
+        ndarray
+            Estimated mean.
+        """
+        pass
```
umarteauowkin, line 46: Ridiculous comment: not sure pass is needed (since there is a docstring)

pydeseq2/dds.py
```diff
@@ -707,31 +637,19 @@ def fit_LFC(self) -> None:
...
-                    counts=self.X[:, i],
-                    size_factors=self.obsm["size_factors"],
-                    design_matrix=design_matrix,
-                    disp=self.varm["dispersions"][i],
-                    min_mu=self.min_mu,
-                    beta_tol=self.beta_tol,
-                )
-                for i in self.non_zero_idx
-            )
+        MLE_lfcs_, mu_, hat_diagonals_, converged_ = self.inference.irls(
```
umarteauowkin, line 640: MLE -> mle for case convention ?

umarteauowkin (approved):
Great  PR, really brings readability, and will be super useful to devise custom inference routines :) Thanks a lot ! Two very very minor linting suggestions, but feel free to ignore

BorisMuzellec (approved):
Thanks a lot @adamgayoso for this PR, and sorry for the long wait! I'm finally merging this.

PR #215, opened 2023-11-28
[bugfix] use self._backend instead of 'loky' directly

adamgayoso:


BorisMuzellec (approved):
Thanks for catching this

PR #236, opened 2024-02-13
MAINT Remove statsmodels

### LINKED ISSUE #231: [BUG] Negative coefficients in the trend curve
**Describe the bug**
`test_zero_inflated()` (from `test_edge_cases.py`) is currently failing due to coefficients in the trend curve being negative. This causes negative `fitted_dispersions` causing, in turn, bugs when taking the log.

**To Reproduce**
Run `test_zero_inflated()` (from `test_edge_cases.py`)

**Expected behavior**
When this happens, DESeq2 raises a warning and runs a local trend fit, which is not implemented in pydeseq2 yet. The ideal solution would be to implement a python equivalent of [`localDispersionFit`](https://github.com/thelovelab/DESeq2/blob/6947d5bc629015fb8ffb2453a91b71665a164483/R/core.R#L2183) and run it whenever the parametric fit fails.

**Additional context**
[Link](https://github.com/thelovelab/DESeq2/blob/devel/R/core.R#L874-L882) to the relevant DESeq2 R code.

BorisMuzellec:
#### Reference Issue or PRs

(Partially) fixes #231, closes #235

This PR removes pydeseq2's dependence on statsmodels, which has been causing a few bugs recently (and is the reason why the CI currently fails).

Statsmodels was used to:

- fit a gamma GLM in `fit_dispersion_trend` 
    - -> this PR replaces it with a custom implementation based on `scipy.optimize`'s L-BFGS-B (which incidentally solves the negative coefficients issue #231). When the fit fails, the code switches to a mean fit.
- perform multiple test adjustment 
    - -> replaced by [`scipy.stats.false_discovery_control`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.false_discovery_control.html)
- perform lowess (locally weighted regression) in `DeseqStats._independent_filtering` 
    - -> replaced by a custom implementation adapted from @agramfort's [lowess gist](https://gist.github.com/agramfort/850437?permalink_comment_id=3437921)

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -601,51 +601,55 @@ class DeseqDataSet(ad.AnnData):
         # Initialize coefficients
         old_coeffs = pd.Series([0.1, 0.1])
         coeffs = pd.Series([1.0, 1.0])
-        try:
-            while (coeffs > 0).all() and (
-                np.log(np.abs(coeffs / old_coeffs)) ** 2
-            ).sum() >= 1e-6:
-                old_coeffs = coeffs
-                coeffs, predictions = self.inference.dispersion_trend_gamma_glm(
-                    covariates, targets
-                )
-                # Filter out genes that are too far away from the curve before refitting
-                pred_ratios = (
-                    self[:, covariates.index].varm["genewise_dispersions"] / predictions
-                )
-
-                targets.drop(
-                    targets[(pred_ratios < 1e-4) | (pred_ratios >= 15)].index,
-                    inplace=True,
-                )
-                covariates.drop(
-                    covariates[(pred_ratios < 1e-4) | (pred_ratios >= 15)].index,
-                    inplace=True,
-                )
-            self.uns["trend_coeffs"] = pd.Series(coeffs, index=["a0", "a1"])
-
-            self.varm["fitted_dispersions"] = np.full(self.n_vars, np.NaN)
-            self.uns["disp_function_type"] = "parametric"
-            self.varm["fitted_dispersions"][self.varm["non_zero"]] = self.disp_function(
-                self.varm["_normed_means"][self.varm["non_zero"]]
+        while (coeffs > 0).all() and (
+            np.log(np.abs(coeffs / old_coeffs)) ** 2
+        ).sum() >= 1e-6:
+            old_coeffs = coeffs
+            coeffs, predictions, converged = self.inference.dispersion_trend_gamma_glm(
+                covariates, targets
             )
 
-        except RuntimeError:
-            warnings.warn(
-                "The dispersion trend curve fitting did not converge. "
-                "Switching to a mean-based dispersion trend.",
-                UserWarning,
-                stacklevel=2,
-            )
-            self.uns["mean_disp"] = trim_mean(
-                self.varm["genewise_dispersions"][
-                    self.varm["genewise_dispersions"] > 10 * self.min_disp
-                ],
-                proportiontocut=0.001,
+            if not converged:
+                warnings.warn(
+                    "The dispersion trend curve fitting did not converge. "
+                    "Switching to a mean-based dispersion trend.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                self.uns["mean_disp"] = trim_mean(
+                    self.varm["genewise_dispersions"][
+                        self.varm["genewise_dispersions"] > 10 * self.min_disp
+                    ],
+                    proportiontocut=0.001,
+                )
+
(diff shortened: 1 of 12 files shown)


pydeseq2/utils.py
```diff
@@ -1520,3 +1522,73 @@ def make_MA_plot(
...
+):
+    """Run lowess smoothing: Robust locally weighted regression.
+
+    The lowess function fits a nonparametric regression curve to a scatterplot.
+    The arrays features and targets contain an equal number of elements; each pair
+    (features[i], targets[i]) defines a data point in the scatterplot. The function
+    returns the estimated (smooth) values of targets.
+    The smoothing span is given by f. A larger value for f will result in a
+    smoother curve. The number of robustifying iterations is given by iter. The
+    function will run faster with a smaller number of iterations.
```
umarteauowkin, line 1545: Maybe specify that the features and labels array are assumed to be flat and of the same shape

pydeseq2/utils.py
```diff
@@ -1520,3 +1522,73 @@ def make_MA_plot(
...
+    ndarray
+        Estimated (smooth) values of targets.
+    """
+    n = len(features)
+    r = int(ceil(frac * n))
+    h = np.maximum(
+        np.array([np.sort(np.abs(features - features[i]))[r] for i in range(n)]), 1e-12
+    )
+    w = np.clip(
+        np.abs(np.nan_to_num((features[:, None] - features[None, :]) / h)), 0.0, 1.0
```
umarteauowkin, line 1569: maybe for coherence put h[:,None]

pydeseq2/utils.py
```diff
@@ -1520,3 +1522,73 @@ def make_MA_plot(
...
+
+
+# Authors: Alexandre Gramfort <alexandre.gramfort@telecom-paristech.fr>
+#
+# License: BSD (3-clause)
+
+# Adapted from https://gist.github.com/agramfort/850437
+
+
+def lowess(
```
umarteauowkin, line 1534: LGTM

pydeseq2/dds.py
```diff
@@ -477,7 +477,7 @@ def fit_size_factors(
             warnings.warn(
                 "Every gene contains at least one zero, "
                 "cannot compute log geometric means. Switching to iterative mode.",
-                RuntimeWarning,
+                UserWarning,
```
umarteauowkin, line 480: Out of curiosity, what is the rationale of selecting UserWarning ?

umarteauowkin (approved):
Looks perfect, I checked the math in the lowess computation and compared to the statsmodel implem, it is the same. Thanks a lot @BorisMuzellec

pydeseq2/dds.py
BorisMuzellec: The bad reason: it was easier to catch it simultaneously with a switch to "mean" trend curve fitting in the tests. The better reason is that I felt it was more suited (`RuntimeWarning` is supposed to cover [dubious runtime behavior](https://docs.python.org/3/library/exceptions.html#RuntimeWarning) which is not really what's going on here).

pydeseq2/utils.py
BorisMuzellec: Do you think it would make the code clearer? Given that the code is working as is, I'm a bit reluctant to add unnecessary broadcasting operations.

pydeseq2/utils.py
BorisMuzellec: Thanks a lot for the review!

pydeseq2/default_inference.py
```diff
@@ -207,17 +202,34 @@ def wald_test(  # noqa: D102
...
+            jac=grad,
+            method="L-BFGS-B",
+            bounds=[(0, np.inf)],
         )
-        res = glm_gamma.fit()
-        coeffs = res.params
-        return (coeffs, covariates_fit @ coeffs)
+
+        if not res.success:
+            raise RuntimeError("Gamma GLM optimization failed.")
```
adamgayoso, line 229: I think it would be cleaner to return `res.success` and raise the error in the while loop

umarteauowkin:
@adamgayoso  From what I gather lowess is not exactly loess, the weighting scheme is automatic in lowess and hand made in loess. @BorisMuzellec correct me if I m wrong, but in scikit-misc, I think the weighted version is not implemented.

adamgayoso:
@umarteauowkin it looks like you can set the weights. Nonetheless I would consider wrapping this custom lowess with a numba njit decorator for speed, as well as test it against the statsmodels one

PR #239, opened 2024-02-19
BUG Fix new all zero replace

### LINKED ISSUE #237: [BUG] KeyError in cooks replacement
**Describe the bug**
KeyError from outlier replacement

**To Reproduce**
```
import pydeseq2
from pydeseq2.utils import load_example_data
from pydeseq2.dds import DeseqDataSet
from pydeseq2.ds import DeseqStats

print(f'version == {pydeseq2.__version__}')
counts_df = load_example_data(
    modality="raw_counts",
    dataset="synthetic",
    debug=False,
)

metadata = load_example_data(
    modality="metadata",
    dataset="synthetic",
    debug=False,
)
metadata=metadata.loc[[f'sample{i}' for i in [*range(1,11),*range(91,101)]]]
counts_df=counts_df.loc[metadata.index]

counts_df.loc[:,'geneX']=0
counts_df.loc['sample100','geneX']=100

dds = DeseqDataSet(
    counts=counts_df,
    metadata=metadata,
    design_factors="condition",
    refit_cooks=True,
)
dds.deseq2()

stat_res = DeseqStats(dds)
stat_res.summary()
```

>version == 0.4.4

>  ...[snipped traceback]
>     583     cooks_outlier = (
> --> 584         (self.dds[use_for_max, :].layers["replace_cooks"] > cooks_cutoff)
>     585         .any(axis=0)
>     586         .copy()
>     587     )
>     589 else:
>     590     cooks_outlier = (
>     591         (self.dds[use_for_max, :].layers["cooks"] > cooks_cutoff)
>     592         .any(axis=0)
>     593         .copy()
>     594     )
> 
> KeyError: 'replace_cooks'
>

### LINKED ISSUE #238: Why is replacement not done if it would result in all zero counts?
[here](https://github.com/owkin/PyDESeq2/blob/39b6a373abb85991b5ac50f5f5b26a1a290d890b/pydeseq2/dds.py#L965C1-L966C2) and [here](https://github.com/owkin/PyDESeq2/blob/39b6a373abb85991b5ac50f5f5b26a1a290d890b/pydeseq2/dds.py#L1023-L1024) there is a filter for not doing outlier replacement or refitting if the result of the replacement would be all zero counts.
```
        # Only refit genes for which replacing outliers hasn't resulted in all zeroes
        new_all_zeroes = (self.counts_to_refit.X == 0).all(axis=0)
```

```
        # Only replace if genes are not all zeroes after outlier replacement
        to_replace[to_replace] = ~new_all_zeroes
```

Why is this the right thing to do? If I have zero counts in all but one sample, and a large count in the single outlier sample, why wouldn't the right result be to remove the outlier?

Is this consistent with what the R DESeq2 package does?

BorisMuzellec:
#### Reference Issue or PRs

Fixes #237, closes #238

This PR fixes a bug arising when all genes whose Cooks outliers are replaced lead to all-zero values (cf #237).

To fix this, it introduces (similarly to [DESeq2](https://github.com/thelovelab/DESeq2/blob/0247fd7b814c47853891aefe06a127721927333d/R/core.R#L2492)) a new `varm` field, `"refitted"`, which complements `"replaced"` by indicating whether dispersions and LFCs were refitted for that gene (on top of its having values replaced).

Finally, the faulty example presented in #237 was added to the test suite.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -601,7 +601,7 @@ class DeseqDataSet(ad.AnnData):
         # Initialize coefficients
         old_coeffs = pd.Series([0.1, 0.1])
         coeffs = pd.Series([1.0, 1.0])
-        while (coeffs > 0).all() and (
+        while (coeffs > 1e-10).all() and (
             np.log(np.abs(coeffs / old_coeffs)) ** 2
         ).sum() >= 1e-6:
             old_coeffs = coeffs
@@ -609,7 +609,7 @@ class DeseqDataSet(ad.AnnData):
                 covariates, targets
             )
 
-            if not converged:
+            if not converged or (coeffs <= 1e-10).any():
                 warnings.warn(
                     "The dispersion trend curve fitting did not converge. "
                     "Switching to a mean-based dispersion trend.",
@@ -628,7 +628,12 @@ class DeseqDataSet(ad.AnnData):
                     self.n_vars, self.uns["mean_disp"]
                 )
 
-                break
+                end = time.time()
+
+                if not self.quiet:
+                    print(f"... done in {end - start:.2f} seconds.\n", file=sys.stderr)
+
+                return
 
             # Filter out genes that are too far away from the curve before refitting
             pred_ratios = (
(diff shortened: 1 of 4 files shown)


umarteauowkin (approved):
LGTM, thanks a lot @BorisMuzellec

PR #243, opened 2024-02-26
BUG fix independent filtering

### LINKED ISSUE #242: IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed.
Hello,
I'm trying to conduct a multifactor analysis on my data. I completely prepared the data in accordance with the tutorial, and at the stage of statistical analysis (.summary() function) I received the error “too many indices for array: array is 1-dimensional, but 2 were indexed”. Moreover, if I reduce the number of genes from 30 thousand to 10 genes, no error occurs. Could you tell me what is the reason?

BorisMuzellec:
#### Reference Issue or PRs

Fixes #242 

This PR fixes a bug in `DeseqStats._independent_filtering` caused by lowess. More precisely:

- we remove the extra index, no longer required as the output of lowess is now 1-dimensional
- we fix the order of the arguments, which was inverted
- we handle the rare case in lowess where the median of residuals is 0

As this bug was not caught by the CI, we add a test case on a synthetic dataset with 50 genes, on which we compare the output of PyDESeq2 with those of the R package DESeq2. In particular, this test case fails without the above fixes.

### DIFF
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -535,7 +535,6 @@ class DeseqStats:
         else:
             residual = num_rej[num_rej > 0] - lowess_res[num_rej > 0]
             thresh = lowess_res.max() - np.sqrt(np.mean(residual**2))
-            print(thresh)
             if np.any(num_rej > thresh):
                 j = np.where(num_rej > thresh)[0][0]
             else:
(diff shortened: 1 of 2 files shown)


umarteauowkin (approved):
Thanks a lot @BorisMuzellec  for this PR, looks good to me !

PR #266, opened 2024-04-02
Fix: latex display in sphinx doc

tanguy-marchand:

- Fix the rendering of latex formula in docstring when making html doc with sphinx, by transforming all "\\" into "\".

I checked the result by rebuilding the doc and checking that the formulae are properly rendered.
Before:
image

After:
image

- I did not check whether the formulae were right, but just that they were well-displayed.
--
When compiling the doc, I get the following warning due to cycling import:

```
WARNING: autodoc: failed to import module 'grid_search' from module 'pydeseq2'; the following exception was raised:
cannot import name 'grid_fit_alpha' from partially initialized module 'pydeseq2.grid_search' (most likely due to a circular import) (/Users/owkin/Project/PyDESeq2/pydeseq2/grid_search.py)
```

Indeed `grid_search` imports `utils` and vice versa.  But I don't think it should be solved in this PR.

### DIFF
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -468,7 +468,7 @@ class DeseqStats:
 
         Useful for looking at log fold-change versus mean expression
         between two groups/samples/etc.
-        Uses matplotlib to emulate make_MA() function in DESeq2 in R.
+        Uses matplotlib to emulate the ``make_MA()`` function in DESeq2 in R.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


umarteauowkin:
Capture d’écran 2024-04-02 à 10 54 48
While we're at it, this formula is wrong, the second y_i should be a mu_i :)

umarteauowkin (changes requested):
I requested a change to correct an error in the formula, but otherwise LGTM. Thanks !

tanguy-marchand:
> Capture d’écran 2024-04-02 à 10 54 48 While we're at it, this formula is wrong, the second y_i should be a mu_i :)

New formula:
image

umarteauowkin (approved):
Perfect, thanks :)

Issue #264, opened 2024-04-01
support similar apeAdapt argument as in DESeq2::lfcShrink()


awalsh17:
**Is your feature request related to a problem? Please describe.**
This is not a problem, but a limitation for some datasets.
In the R package, the DESeq2::lfcShrink() wrapper that calls {apeglm} includes the `apeAdapt` argument that, when set to FALSE, does not use the logFC MLE and sets mle = NULL for apeglm() call. This thread explains some of the reasoning to do this: https://support.bioconductor.org/p/111544/ 

**Describe the solution you'd like**
Would you support adding an argument to `lfc_shrink` that would perform similarly to the R implementation? 
If you pointed me in the right direction in the code, I could implement it. 
If it is as simple as changing `prior_scale` to 1, then this might be simple. (Based on my read of the code here: https://github.com/azhu513/apeglm/blob/master/R/apeglm.R)

**Describe alternatives you've considered**
Just use the R package 

**Additional context**
Thank you!

PR #267, opened 2024-04-02
Add adapt = False (prior_scale=1) option to lfc_shrink()

awalsh17:
#### Reference Issue or PRs

This is in reference to #264 

This adds an option to lfc_shrink() to allow one to keep prior_scale = 1. 
This is equivalent to the R package implementation of DESeq2::lfcShrink( apeAdapt = FALSE )

I tested some and the behavior is similar to the R package. But not 100%. I am not sure, but likely a difference in the parameters I passed to each. 
These plots are from using the same data, contrast, and adapt = False argument.

image


BorisMuzellec:
Thanks a lot for this PR @awalsh17! I'll try to add a test to automatically compare with deseq2 by the end of the week.

BorisMuzellec:
> I tested some and the behavior is similar to the R package. But not 100%. I am not sure, but likely a difference in the parameters I passed to each. These plots are from using the same data, contrast, and adapt = False argument.

@awalsh17 in those plots did you run lfc shrinkage starting from the same size factors, dispersions and LFCs, or did you run a DESeq2 pipeline and a PyDESeq2 pipeline independently (starting from the same data)?

awalsh17:
> > I tested some and the behavior is similar to the R package. But not 100%. I am not sure, but likely a difference in the parameters I passed to each. These plots are from using the same data, contrast, and adapt = False argument.
> 
> @awalsh17 in those plots did you run lfc shrinkage starting from the same size factors, dispersions and LFCs, or did you run a DESeq2 pipeline and a PyDESeq2 pipeline independently (starting from the same data)?

@BorisMuzellec Good point. I ran them completely independently (starting from the same data). I could repeat using the same data input to just the lfc shrinkage?

BorisMuzellec:
> @BorisMuzellec Good point. I ran them completely independently (starting from the same data). I could repeat using the same data input to just the lfc shrinkage?

If you have time for it, a good sanity check would be to test on real data indeed, but for this we would need to factor out differences between DESeq2 and PyDESeq2 in terms of dispersions by manually setting the same dispersions in both packages.

Otherwise, I think that the PR passing tests on the synthetic test data is good enough

BorisMuzellec (approved):

PR #268, opened 2024-04-09
fix : enable vst be used "blindly" and to fit its own dispersion

### LINKED ISSUE #263: [BUG] The vst() function does not work when the algorithm switches to a mean-based dispersion trend.
**Describe the bug**
Calling the vst() leads to the fit_type variable being hard-coded to the chosen fit_type. Consequently when the vst_transform is called it breaks with:
`
File ".venv/lib/python3.10/site-packages/pydeseq2/dds.py", line 411, in vst_transform
    a0, a1 = self.uns["trend_coeffs"]
KeyError: 'trend_coeffs'
`

**Expected behavior**
When switching to mean based, update fit_type to 'mean' in line 918.
`
self.fit_type = 'mean'
`

laudmt:
#### Reference Issue or PRs

Fix #263

Proposed solution :

- A `self.fit_type` is defined at class initialisation and will by default set the fit_type for DEA and VST.
- If needed, a fit type can be passed to deseq2() and vst() in order to launch the two analysis with a separate fit_type. It will set self.fit_type to the user one.
- self.vst_fit() and self.vst_transform() will call internally the self.fit_type.
- vst_transform() should always be called after vst_fit() or deseq2(). Raise an exception if needed trend_coef have not been computed by deseq2() or vst_fit()
- Add unit tests

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -311,7 +311,8 @@ class DeseqDataSet(ad.AnnData):
             If False, only an intercept is used. (default: ``False``).
 
         fit_type: str
-            * ``None``: use the vst_fit_type to fit the dispersions.
+            * ``None``: fit_type provided at initialization to fit
+            the dispersions trend curve.
             * ``"parametric"``: fit a dispersion-mean relation via a robust
               gamma-family GLM.
             * ``"mean"``: use the mean of gene-wise dispersion estimates.
@@ -458,8 +459,8 @@ class DeseqDataSet(ad.AnnData):
             (default: ``None``).
         """
         if fit_type is not None:
-            self.current_fit_type = fit_type
-            print(f"Using {self.current_fit_type} fit type.")
+            self.fit_type = fit_type
+            print(f"Using {self.fit_type} fit type.")
 
         # Compute DESeq2 normalization factors using the Median-of-ratios method
         self.fit_size_factors()
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -320,8 +320,11 @@ class DeseqDataSet(ad.AnnData):
             (default: ``None``).
         """
         if fit_type is not None:
-            self.fit_type = fit_type
-            print(f"fit type used : {self.fit_type}")
+            self.vst_fit_type = fit_type
+        else:
+            self.vst_fit_type = self.fit_type
+
+        print(f"Fit type used for VST : {self.vst_fit_type}")
 
         self.vst_fit(use_design=use_design)
         self.layers["vst_counts"] = self.vst_transform()
@@ -340,18 +343,29 @@ class DeseqDataSet(ad.AnnData):
         ----------
         use_design : bool
             Whether to use the full design matrix to fit dispersions and the trend curve.
-            If False, only an intercept is used. (default: ``False``).
+            If False, only an intercept is used.
+            Only useful if ``fit_type = "parametric"`.
+            (default: ``False``).
         """
         # Start by fitting median-of-ratio size factors if not already present,
         # or if they were computed iteratively
         if "size_factors" not in self.obsm or self.logmeans is None:
             self.fit_size_factors()  # by default, fit_type != "iterative"
 
+        if not hasattr(self, "vst_fit_type"):
+            self.vst_fit_type = self.fit_type
+
         if use_design:
-            # Check that the dispersion trend curve was fitted. If not, fit it.
-            # This will call previous functions in a cascade.
-            if "disp_function" not in self.uns:
-                self.fit_dispersion_trend()
+            if self.vst_fit_type == "parametric":
+                self._fit_parametric_dispersion_trend(vst=True)
+            else:
+                warnings.warn(
+                    "use_design=True is only useful when fit_type='parametric'. ",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                self.fit_genewise_dispersions(vst=True)
+
         else:
             # Reduce the design matrix to an intercept and reconstruct at the end
             self.obsm["design_matrix_buffer"] = self.obsm["design_matrix"].copy()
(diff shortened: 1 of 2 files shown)


BorisMuzellec:
Thanks for the PR @laudmt!

I think it does the trick for VST, but I have a little concern regarding the fact that providing a `fit_type` other than `None` overwrites a `DeseqDataSet`'s `trend_fit_type`.

Here's an example of problematic / ambiguous behavior that could occur:

```
dds = DeseqDataSet(counts=counts, metadata=metadata, trend_fit_type="parametric")
# Start by computing VST
dds.vst(fit_type="mean")
# Then perform DEA
dds.deseq2() # what trend_fit_type is used to compute the curve ?
```

In the above example, ``dds.deseq2()`` would use `trend_fit_type = "mean"` (because it was overwritten by `vst`), but the user probably intended to fit VST with a curve and then run DEA with a parametric curve.

I think we should find a way to make the choices of curve fit type for `vst()` and `deseq2()` completely independent.

laudmt:
> Thanks for the PR @laudmt!
> 
> I think it does the trick for VST, but I have a little concern regarding the fact that providing a `fit_type` other than `None` overwrites a `DeseqDataSet`'s `trend_fit_type`.
> 
> Here's an example of problematic / ambiguous behavior that could occur:
> 
> ```
> dds = DeseqDataSet(counts=counts, metadata=metadata, trend_fit_type="parametric")
> # Start by computing VST
> dds.vst(fit_type="mean")
> # Then perform DEA
> dds.deseq2() # what trend_fit_type is used to compute the curve ?
> ```
> 
> In the above example, `dds.deseq2()` would use `trend_fit_type = "mean"` (because it was overwritten by `vst`), but the user probably intended to fit VST with a curve and then run DEA with a parametric curve.
> 
> I think we should find a way to make the choices of curve fit type for `vst()` and `deseq2()` completely independent.

Indeed thank you, I was missing this use case.
I will then store two different fit type one for dea and one for vst and enable self.fit_X methods to take a fit_type in argument.

pydeseq2/dds.py
```diff
@@ -432,11 +440,27 @@ def vst_transform(self, counts: Optional[np.ndarray] = None) -> np.ndarray:
...
+            dispersions to the mean intensity.``"parametric"``: fit a dispersion-mean
+            relation via a robust gamma-family GLM. ``"mean"``: use the mean of
+            gene-wise dispersion estimates.
+
+            If None, the fit_type provided at class initialization is used.
+            (default: ``None``).
         """
+        if fit_type is not None:
+            self.current_fit_type = fit_type
+            print(f"Using {self.current_fit_type} fit type.")
```
BorisMuzellec, line 462: `self.current_fit_type` doesn't seem to be used elsewhere, so the `deseq2()` `fit_type` argument doesn't have any effect

pydeseq2/dds.py
```diff
@@ -310,37 +309,38 @@ def vst(
...
+            * ``"parametric"``: fit a dispersion-mean relation via a robust
+              gamma-family GLM.
+            * ``"mean"``: use the mean of gene-wise dispersion estimates.
+
+            (default: ``None``).
         """
-        self.vst_fit(use_design=use_design, fit_type=fit_type)
+        if fit_type is not None:
+            self.fit_type = fit_type
+            print(f"fit type used : {self.fit_type}")
```
BorisMuzellec, line 323: If we set a new `fit_type` in `VST`, then I think we should either set it back to the one provided at initialization, or make it clear that it will also change the fit type used in `deseq2()`

pydeseq2/dds.py
```diff
@@ -310,37 +309,38 @@ def vst(
         use_design : bool
             Whether to use the full design matrix to fit dispersions and the trend curve.
             If False, only an intercept is used. (default: ``False``).
+
         fit_type: str
-            Either "parametric" or "mean" for the type of fitting of dispersions to the
-            mean intensity. "parametric": fit a dispersion-mean relation via a robust
-            gamma-family GLM. "mean": use the mean of gene-wise dispersion estimates.
-            (default: ``"parametric"``).
+            * ``None``: use the vst_fit_type to fit the dispersions.
```
BorisMuzellec, line 314:
```suggestion
            * ``None``: use the fit_type provided at initialization to fit the dispersions trend curve.
```

pydeseq2/dds.py
laudmt: If I understand correctly, the user should be able, after he launches vst with its own fit_type, to launch deseq2 and expect the fit_type he provided at initialisation?

Is there a risk if deseq2 and vst have their own separate fit_type (lets say self.vst_fit_type and self.dea_fit_type) and since they both share all the other attributes filled during the fit, that vst will compute stuff that deseq2 will have access to it but is not supposed to?

In that case, should they not share any of the attributes (varm, obs, design_factors etc) ?

pydeseq2/dds.py
laudmt: Updated

pydeseq2/dds.py
BorisMuzellec: Sorry for the late answer @laudmt!

> If I understand correctly, the user should be able, after he launches vst with its own fit_type, to launch deseq2 and expect the fit_type he provided at initialisation?

I think that this is indeed what should happen by default, i.e. if no other `fit_type` is provided when calling `deseq2()`.

> Is there a risk if deseq2 and vst have their own separate fit_type (lets say self.vst_fit_type and self.dea_fit_type) and since they both share all the other attributes filled during the fit, that vst will compute stuff that deseq2 will have access to it but is not supposed to?
> 
> In that case, should they not share any of the attributes (varm, obs, design_factors etc) ?

Currently, this is what happens. `vst()` and `deseq2()` populate the same fields (`varm["genewise_dispersions"]` and `uns["trend_coeffs"]` in particular), but they recompute and overwrite them every time (except in the case when we call `vst(use_design=True)` after calling `deseq2()` with the same fit type, in which case we can re-use the computation).

In principle, the only issue I see with this is if a user calls e.g. `vst(use_design=False)` after `deseq2()`. In that case, `varm["genewise_dispersions"]` will be overwritten, and if the user wants to perform some analysis of the gene-wise dispersions (e.g. if they call `plot_dispersions()`), then they will access the wrong values (not the ones used in the DEA).

Perhaps a cleaner thing to do would be to prefix any parameter stored by `vst()` with `"vst_"`? (`varm["vst_genewise_dispersions"]`, `uns["vst_trend_coeffs"]`, etc.). This would create some memory overhead but would prevent any confusion. What do you think?

BorisMuzellec:
Hi @laudmt, I've pushed a commit in which I implemented the solution I suggested earlier, i.e. setting two separate `fit_type` and `vst_fit_type` attributes and duplicating fields (with or without a `"vst_"` prefix) to avoid data leakage.

Could you have a look when you have time and tell me if that would work for you :) ?

PR #284, opened 2024-05-30
Poscount implementation & diag(XXT) optimization

asistradition:
Thank you for implementing DESeq2 in python. I have been using it for single-cell data.

Included is an implementation of the poscount size_factor method from DESeq2 (R).

Included is an enhancement in `irls_solver` that replaces the full matrix multiplication to get `diag(X(XTX)-1XT)` with a more memory and runtime-efficient calculation of only the diagonal values. The result is identical.


asistradition:
I am also adding an optimization for `wald_test` that replaces the diagonal weight matrix `W` with a broadcast vector multiplication

umarteauowkin (commented):
Hi @asistradition, thanks a lot for this PR, I learned a trick ! Could you please 
- make sure that the pre-commit CI passes (should just be adding a point)
- Remove the make_H function and simply write the four lines of code in the main code instead (no need to create an auxiliary function for this, in order not to increase the depth. Could you add as a comment what computation this operation does so that this remains clear as the function is less explicit than the previous computation. 
- Could you add a description of the new mode you added, as well as a reference, to check that it is "right"?
Thanks a lot for this contribution ! Happy to approve once these changes are made.

asistradition:
Done. Changed the poscounts method here to match DESeq2 1.34 to float precision and added unit tests for fit_size_factors.

umarteauowkin (approved):
Thank you very much for this PR @asistradition, everything is ood for me now !

PR #286, opened 2024-06-06
summary function uses quiet to print

yihming:

Make `DeseqStats.summary()` function also rely on `quiet` parameter for verbose.


umarteauowkin (approved):
LGTM, thanks for spotting this ! If the checks pass, ready to merge.

Issue #288, opened 2024-06-09
Setting different n_cpus in DefaultInference gives results in the same time


yihming:
**Describe the bug**
I tried setting different `n_cpus` value for `DefaultInference` (e.g. 1 and 8):
```python
from pydeseq2.default_inference import DefaultInference
inference = DefaultInference(n_cpus=n_cpus)
```
Then `dds.deseq2()` and `DeseqStats.summary()` give the similar time (~32s) to finish for the same data.

**To Reproduce**
Provide snippets of code and steps on how to reproduce the behavior. 
Please also specify the version you are using.

**Expected behavior**

I wonder if I didn't set the parameter correctly in order to speed up the computation by parallel computing. Or if I did set it correctly, why the calculation did not take effects. Thanks!

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04.4 LTS with 8 vCPUs available
 - Version: pydeseq2 v0.4.9

**Additional context**

abearab:
I can confirm I also experienced the same issue.

umarteauowkin:
Hi ! Thanks for reporting this. Could you tell me the number of genes with which you experience this issue ?

yihming:
Hi @umarteauowkin ,

Thanks for checking this issue. I tried your suggestion, and found that the Inference object's `n_cpus` is always overwritten to use all available vCPUs.

I tried to fix the issue in PR #293 . It works at my side. Please let me know if I need to do anything further. Thanks!

umarteauowkin:
Solved by #293. Thanks @yihming !

PR #293, opened 2024-06-13
Fix Inference n_cpus overwritten issue

yihming:
#### Reference Issue or PRs

This PR is to fix Issue #288 .

The current initialization of `DeseqDataSet` always overwrites its `Inference` object's `n_cpus`. So no matter which value is specified for `n_cpus` when declaring an `Inference` object, when passed to `DeseqDataSet`, its `n_cpus` is always overwritten to use all available vCPUs.

My implementation just restricts that Inference's `n_cpus` can only be overwritten if DeseqDataSet's `n_cpus` is not `None`.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -284,7 +284,8 @@ class DeseqDataSet(ad.AnnData):
 
         if inference:
             if hasattr(inference, "n_cpus"):
-                inference.n_cpus = n_cpus
+                if n_cpus:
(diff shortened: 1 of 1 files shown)


umarteauowkin (commented):
I don't really understand what happens here. Could you describe why you cannot set the number of cpus without this change ? Happy to accept once I understand.

umarteauowkin (approved):

umarteauowkin:
Thanks a lot @yihming :)

PR #294, opened 2024-06-13
DOC: specify what happens if n_cpus and inference are set

umarteauowkin:
#### Reference Issue or PRs

Adds a more precise docstring on what happens for #293 

We add a more precise docstring on how `n_cpus` is overwritten in `inference`


maikia (approved):
LGTM

PR #304, opened 2024-07-31
Respect quiet=True in DeseqStats.lfc_shrink

maltekuehl:
#### What does your PR implement?

Currently, `DeseqStats.lfc_shrink` does not respect `self.quiet` when printing `self.results_df` at the end of the method. This PR adds checks for `self.quiet` for the missing print statements.


BorisMuzellec (approved):
LGTM, thanks @maltekuehl

PR #314, opened 2024-09-16
BUG set vst fit type to mean when parametric fails

### LINKED ISSUE #311: [BUG] vst() throws a runtime error when parametric fitting defaults to mean because vst_fit_type is still set to parametric
vst_transform() runs into the runtime error when parametric fit does not converge and the algorithm picks the mean method.

**Expected behavior**
In vst(), set self.vst_fit_type to mean if parametric falls back to the mean-based method.

 - OS: Ubuntu
 - Version 0.4.10

BorisMuzellec:
#### Reference Issue or PRs
Fixes #311

Sets ``vst_fit_type="mean"`` in case the parametric fit fails in ``vst_fit()``.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -1098,7 +1098,6 @@ class DeseqDataSet(ad.AnnData):
             Whether the dispersion trend curve is being fitted as part of the VST
             pipeline. (default: ``False``).
         """
-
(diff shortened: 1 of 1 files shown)


umarteauowkin (approved):
LGTM

PR #317, opened 2024-09-20
ENH volcano plots

BorisMuzellec:

This PR implements a method in `DeseqStats` to make volcano plots.

Example:

```
counts = load_example_data()
metadata = load_example_data('metadata')

dds = DeseqDataSet(counts=counts, metadata=metadata)
dds.deseq2()

ds = DeseqStats(dds)
ds.summary()

ds.plot_volcano(LFC_threshold = .25, pval_threshold = 0.05)
```

volcano_plot_example

### DIFF
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -509,14 +509,50 @@ class DeseqStats:
 
     def plot_volcano(
         self,
-        LFC_threshold=0.25,
+        LFC_threshold=2.0,
         pval_threshold=0.05,
         annotate_genes=True,
         write_legend=False,
         save_path=None,
         figsize=(6, 6),
     ):
-        # TODO : Add docstring
+        """
+        Create a volcano plot using matplotlib.
+
+        Summarizes the results of a differential expression analysis by plotting
+        the negative log10-transformed adjusted p-values against the log2 fold change.
+
+        Parameters
+        ----------
+        LFC_threshold : float
+            Log2 fold change threshold above which genes are considered differentially
+            expressed. (default: ``2.``).
+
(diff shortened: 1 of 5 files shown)


umarteauowkin (approved):
I approve as it is a good first draft ! In an ideal world, all the fixed kwargs should be modifiable, but in that case I think people would just recopy the code !

mschilli87:
Any reason this never got merged? I'd be happy to have something like this available.

grst:
I'm not sure if including plotting dependencies in pydeseq2 is the way to go (tbf, they could be optional). 

The scverse ecosystem offers plotting functions for differential expression via pertpy: https://pertpy.readthedocs.io/en/stable/tutorials/notebooks/differential_gene_expression.html

Our vision would be to have a separate, method-agnostic package for plotting differential expression results, but this never made it past prototype stage: https://github.com/scverse/deres

mschilli87:
@grst: Fair points. However, as of now, I get MA plots from PyDESeq2 but no Volcano plots, with both of them coomonly used in the field and there is a pull-request that would add this feature in a simpel way while we all wait for the better solution. Who know, accepting this over a year ago, or at least providing this kind of feedback to the (external?) contributor that provided the PR might have motivated them to contribute more to the ecosystem and potentially even implementing this in in the context of deres.
Pertpy is a very heavy dependency (both in transient dependency load and speed of development) for something as simple as a volcano plot, but thank you for the pointers.

I just want to make clear that I am not complaining or trying to be pushy here. I think you all are doing a great job with scverse and I know very well how easy these kinds of things can end up low enough on everyone's priority lists to eventually never materialize.

PR #324, opened 2024-10-17
ENH: Memory Optimizations & low_memory Flag

asistradition:

The `replace_cooks` data was stored as a pandas DataFrame. This is memory inefficient for large data. It has been refactored from a DataFrame to a numpy array.

A new `low_memory` argument to `DeseqDataSet` is available. When set to True, large data arrays saved into the AnnData elements `.obsm` and `.layers` are deleted if there is no further use for them in the standard deseq workflow. This reduces peak memory consumption dramatically.

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -290,9 +290,6 @@ class DeseqDataSet(ad.AnnData):
         self.logmeans = None
         self.filtered_genes = None
 
-        # Cooks outliers
-        self._cooks_outlier = None
-
         if inference:
             if hasattr(inference, "n_cpus"):
                 if n_cpus:
@@ -974,8 +971,8 @@ class DeseqDataSet(ad.AnnData):
 
     def cooks_outlier(self):
         """Filter p-values based on Cooks outliers."""
-        if self._cooks_outlier is not None:
-            return self._cooks_outlier
+        if '_pvalue_cooks_outlier' in self.varm.keys():
+            return self.varm['_pvalue_cooks_outlier']
 
         num_samples = self.n_obs
         num_vars = self.obsm["design_matrix"].shape[-1]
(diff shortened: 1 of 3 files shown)


pydeseq2/dds.py
```diff
@@ -115,6 +115,10 @@ class DeseqDataSet(ad.AnnData):
     quiet : bool
         Suppress deseq2 status updates during fit.
 
+    low_memory : bool
+        Remove intermediate data structures from .layers and from .obsm that are no
+        longer necessary after they are used during deseq2 run. (default: False)
```
BorisMuzellec, line 120:
```suggestion
        longer necessary after they are used during deseq2 run, such as Cook's
        distances. (default: False)
```

pydeseq2/dds.py
```diff
@@ -281,9 +286,13 @@ def __init__(
         self.min_replicates = min_replicates
         self.beta_tol = beta_tol
         self.quiet = quiet
+        self.low_memory = low_memory
         self.logmeans = None
         self.filtered_genes = None
 
+        # Cooks outliers
+        self._cooks_outlier = None
```
BorisMuzellec, line 294: Following the anndata logic, It seems like it would be more natural to store this in `.varm` rather than as a `DeseqDataSet` attribute on its own.

I would also suggest renaming to `_pvalue_cooks_outlier` to disambiguate between genes whose counts were replaced based on cook's distances (`.varm['replaced']`) and those whose p-values are set to NaN later on (again based on cook's).

BorisMuzellec (changes requested):
Hi @asistradition, thanks a lot for this PR!

Overall I agree with the changes you propose, I just have a few comments regarding the storage and naming of `._cooks_outlier` (cf above).

Just for the sake of curiosity, were you limited in your usage of PyDESeq2 because of memory consumption? If so, I'd be curious to have an idea of your setup and the size of your dataset :)

pydeseq2/dds.py
asistradition: Moved the outliers mask into `self.varm['_pvalue_cooks_outlier']`

asistradition:
Yes, this is part of my standard single-cell workflow, and memory limitations are the main bottleneck. 60k x 35k won't be a problem with this PR.

BorisMuzellec (approved):
Perfect, thanks!

PR #181, opened 2023-10-17
Adding interaction terms to the design matrix

khalilouardini:
This PR has several purposes:
- it leverages [formulaic]((https://matthewwardrop.github.io/formulaic/guides/contrasts/#treatment-aka-dummy)) to allow using interaction terms of the form `a:b:...:z`for `design_factors` and formulas based on a combination of single design factors and of such interaction terms `a + b + a:b:...:z` (no support for other more complex structures or alternative syntax enabled by formulaic's grammar yet such as i.e. `~ C(X, contr.treatment("x"))`, `a * b`, `contr.poly`, ...)
- allows the user to pass a `design_matrix`to a dds dataset
- fixes `pyproject.toml` syntax (linters complained)

Completion milestones:
- [x] adds @anaischossegros test on edge-case from deseq2 vignette
- [x] test to check that passing a design_matrix populates all necessary attributes such as `design_factors` (kind of as end2end integration tests are passing)
- [x] continuous factors well-handled
- [x] deseq2 examples matched (at least at the level of the design matrix)
- [x] arbitrary number of interactions well handled
- [x] deseq2 examples matched end2end

PR #328, opened 2024-10-29
ENH Generic design support using `formulaic`

### LINKED ISSUE #125: Use formulaic for dealing with design matrices
Have y'all considered using [formulaic](https://github.com/matthewwardrop/formulaic) for handling your design matrices? (The successor to patsy.) It may simplify a bunch of work for supporting more complex experiment designs.

(Only just learned of this project btw...looks really great!)

### LINKED ISSUE #184: [Enhancement] `expanded` argument not used in `utils.build_design_matrix`
**Describe the bug**
the expanded argument in  `utils.build_design_matrix`  is not used

**Additional context**
related to [PR#181](https://github.com/owkin/PyDESeq2/pull/181)

### LINKED ISSUE #202: Explanation of ref_level argument of DeseqDataSet() in the tutorials
Hello,
While contrast argument of DeseqStats() is explained well in "A simple PyDESeq2 workflow", ref_level argument of DeseqDataSet() and its importance to match the contrasts in design_matrix and DeseqStats are not described. Although the error message directly suggests a solution, seeing it beforehand in the documentation would be better.
Thank you for the tool. Regards.

### LINKED ISSUE #213: Support for arbitratry design matrices and contrast vectors
**Is your feature request related to a problem? Please describe.**
Most linear models support passing designs as design matrices and contrasts as contrast vectors. This is the "smallest common denominator" for specifying designs and it's useful
 * for more complex designs and comparisons that aren't covered by a simple `[column, baseline, treatment]` triplet
 * for writing wrapper functions (e.g. [multi-condition-comparisions](https://github.com/scverse/multi-condition-comparisions))  that use PyDESeq2 as one of multiple backends and already deal with building model matrices and contrast vectors from more user-friendly input such as formulae. 

**Describe the solution you'd like**
 * `DeSeqDataset` should take a design matrix
 * `DeseqStats` should take a contrast vector with one value per fitted coefficient, such as `[0, -1, 1]`. 

**Additional context**
discussed on the scverse hackathon in Cambridge

CC @const-ae @emdann

### LINKED ISSUE #272: Simplify naming convention conversion between pydeseq2 and formulaic
This information is already available from formulaic directly. When `design` is your design matrix, there is
```
design.model_spec.variable_terms
```
and 
```
design.model_spec.term_variables
```

to map from variables to formulae terms and vice versa.

_Originally posted by @grst in https://github.com/owkin/PyDESeq2/pull/181#discussion_r1549442869_

The complex logic used to handle interaction terms could indeed be simplified (as long as this logic is still needed).

### LINKED ISSUE #309: Feature request: test for continuous variables and create design matrix from formula
Dear all, 
thank you for the great package, it´s a staple in my workflows.

Recently, I had the task to analyze continuous covariates in my RNA-seq data. I know continuous variables can be included in the model, but I would like to estimate the p-values for the logfoldchanges, and the `DESeqStats` class takes only a contrast, e.g. from a categorical factor. 
Estimating continuous variables would be amazing.
(I did check and test if it´s possible, but could not find anything. Please correct me if I am wrong.)

A minor feature request: Add the option to create the design from a mode formula (`patsy`-style) in `DESeqDataSet`. If I can set the reference level of several categorical factors in my data and then create the design matrix from a formula (in comparison to setting the `ref_level` of only one design factor), I could fit a model once and then extract `DESeqStats` for several variables.

Thank you for any help and best regards,
Max

BorisMuzellec:
#### Reference Issue or PRs
Closes #181
Closes #213 
Closes #309
Closes #272
Closes #202 
Closes #184 
Closes #125
Will unblock scverse/pertpy#610

This PR implements support for general design matrices thanks to [formulaic](https://github.com/matthewwardrop/formulaic), and using utils from [pertpy](https://github.com/scverse/pertpy).

#### `DeseqDataSet`

* Designs should now be provided to `DeseqDataSet` using the `design` argument, either in the form of a string representing a `formulaic` formula (e.g. `"~condition + treatment"`, `"~condition + condition:treatment"`, `"~condition + exp(cofactor)"`...), or an ndarray directly corresponding to a design matrix.
* `design_factors` is still supported but throws a `DeprecationWarning`
* `continuous_factors` is deprecated, as continuous type inference is handled by `formulaic`
* `ref_level` is deprecated
* Due to new decorated methods, `DeseqDataSet` is no longer picklable. A `to_picklable_anndata()` method was added to allow users to pickle results for later use.

#### `DeseqStats`

* Default contrasts are no longer supported, as they lead to too many errors
* Contrasts may be provided as before for categorical variables (e.g. `["treatment", "test", "control"]`), or directly in the form of a contrast vector (a numpy array).
* For now, contrasts for continuous variables are directly specified with a contrast vector.
* `lfc_shrink` no longer supports a default `coef` argument

BREAKING CHANGE: python 3.9 is no longer supported.

TODO: 
- [x] Update the example notebook gallery on RTD.
  - [x] ~Failure seems to be due to the fact that it's not possible to pickle classes with decorated functions.~ Solved using `to_picklable_anndata`. 
- [x] Add more tests
  - [x] Port the `_formulaic.py` tests from pertpy 
  - [x] Test inputing a design matrix directly (and not as a formula).


pydeseq2/_formulaic.py
```diff
@@ -0,0 +1,278 @@
...
+from formulaic import FactorValues
+from formulaic import ModelSpec
+from formulaic.materializers import PandasMaterializer
+from formulaic.materializers.types import EvaluatedFactor
+from formulaic.parser.types import Factor
+from interface_meta import override
+
+
+@dataclass
+class FactorMetadata:
```
grst, line 28: There's also quite a bunch of test cases for the `_formulaic.py` file and the `LinearModelBase` in pertpy. Would be great if you could also port them here!

pydeseq2/dds.py
```diff
@@ -58,18 +62,22 @@ class DeseqDataSet(ad.AnnData):
         DataFrame containing sample metadata.
         Must be indexed by sample barcodes.
 
-    design_factors : str or list
-        Name of the columns of metadata to be used as design variables.
-        (default: ``'condition'``).
+    design : str or pandas.DataFrame
```
grst, line 65: not related to this PR, but since you are using typehints anyway, you could rely on [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/) and not have to repeat the type information in the docstrings.

pydeseq2/dds.py
```diff
@@ -226,55 +235,57 @@ def __init__(
...
-        if np.any(["_" in factor for factor in self.design_factors]):
+        if continuous_factors is not None:
             warnings.warn(
-                """Same factor names in the design contain underscores ('_'). They will
-                be converted to hyphens ('-').""",
-                UserWarning,
+                "continuous_factors is deprecated and will soon be removed."
+                "Continuous factors are now automatically detected from the design,"
+                "or can be cast to categorical using the C() operator in the formula",
+                DeprecationWarning,
```
grst, line 247: For backwards compatibility, consider adding something like

```python
design_factors.extend(
    [f"C({f})" for f in continuous_factors]
)
```

pydeseq2/dds.py
```diff
@@ -1016,6 +1037,34 @@ def cooks_outlier(self):
...
+            varm=self.varm,
+            uns=self.uns,
+            layers=self.layers,
+        )
+
+        # Convert the design matrix to a DataFrame to remove model_spec
+        adata.obsm["design_matrix"] = pd.DataFrame(adata.obsm["design_matrix"])
+
+        return adata
+
```
grst, line 1067: As mentioned on zulip, I think a preferred way would be to implement a 
`.save()` and (static) `.load()` method that saves the anndata as h5ad and the state of the model as pickle or any other suited format. 

The `.load()` should be able to restore the full state of the dds object, including the formulaic metadata ideally.

pydeseq2/dds.py
```diff
@@ -305,6 +315,17 @@ def __init__(
...
+    @property
+    def variables(self):
+        """Get the names of the variables used in the model definition."""
+        try:
+            return self.obsm["design_matrix"].model_spec.variables_by_source["data"]
+        except AttributeError:
+            raise ValueError(
+                """Retrieving variables is only possible if the model was initialized
+                using a formula."""
+            ) from None
```
grst, line 327: Maybe this stuff could really become a Mixin as you suggested, then we can more easily reuse it across pyDESeq2 and pertpy.

pydeseq2/ds.py
```diff
@@ -166,16 +164,26 @@ def __init__(
...
+        # Check the validity of the contrast (if provided) or build it.
+        if contrast is None:
+            raise ValueError(
+                """Default contrasts are no longer supported.
+                The "contrast" argument must be provided."""
+            )
+        elif isinstance(contrast, np.ndarray):
+            assert (
+                contrast.shape[0] == self.dds.obsm["design_matrix"].shape[1]
+            ), "The contrast vector must have the same length as the design matrix."
```
grst, line 181: IMO this should raise a ValueError. 

assertions are only meant for internal consistency checks, and might be skipped depending on the python optimization settings.

pydeseq2/ds.py
```diff
@@ -681,31 +585,61 @@ def _build_contrast_vector(self) -> None:
...
-            new_ref_idx = self.LFC.columns.get_loc(f"{factor}_{ref}_vs_{old_ref}")
-            self.contrast_vector[new_alternative_idx] = 1
-            self.contrast_vector[new_ref_idx] = -1
+        for var in self.dds.variables:
+            if var in cond_dict:
+                self.dds._check_category(var, cond_dict[var])
+            else:
+                cond_dict[var] = self.dds._get_default_value(var)
+        df = pd.DataFrame([kwargs])
+        return self.dds.obsm["design_matrix"].model_spec.get_model_matrix(df).iloc[0]
```
grst, line 645: Depends on you if you want to adopt the `.cond()` syntax for building contrasts (which was originally devised by @const-ae in [glmGamPoi](https://www.bioconductor.org/packages/release/bioc/vignettes/glmGamPoi/inst/doc/pseudobulk.html)). 

A *lot* of the code in the `_formulaic.py` module is just around finding the baseline level for each condition such that this works nicely in the case of interaction terms. 

In case you were just to support `[column, baseline, treatment]` and numpy array contrasts, you could probably come up with a way simpler solution.

grst (commented):
I really prefer this over the old appraoch, many thanks for moving this forward @BorisMuzellec!

pydeseq2/dds.py
BorisMuzellec: Not sure how to ensure backwards compatibility, because pydeseq2 currently considers all factors as categorical unless passed to `continuous_factors`, whereas `formulaic` automatically detects continuous (numerical) factors but implements the `C()` operator to force categorical encoding. In a way, these are opposite behaviours.

The closest I can think of would be something like 
```
self.design_factors = [f"{f}" if f in continuous_factors else f"C({f})" for f in design_factors]
```
but that doesn't ensure that factors listed in `continuous_factors` are indeed treated as numerical, only that those which aren't are treated as categorical.

pydeseq2/dds.py
BorisMuzellec: I think that's a good suggestion, but it should probably be treated in a different PR to avoid bloating this one (which is already pretty big).

Also as discussed on Zulip, I think that inheriting from the `AnnData` class wasn't the best choice, and that `DeseqDataSet` should have an `AnnData` component instead. But again, another PR :)

pydeseq2/dds.py
grst: my bad, I think you can just add them to the design factors.

pydeseq2/ds.py
BorisMuzellec: I don't have a strong opinion on this, whatever offers the most flexibility is best. 

At first I tried simplifying the code in `_formulaic.py` to keep only what I need (mainly retrieving levels for a given factor + whether it has numerical or categorical type), but I ended up keeping everything because I didn't find a straightforward simplification.

> A lot of the code in the _formulaic.py module is just around finding the baseline level for each condition such that this works nicely in the case of interaction terms.

How would you define a contrast to test interaction terms using `.cond()`? Right now I don't see how to do it without using a numerical factor.

pydeseq2/dds.py
BorisMuzellec: If it's simpler for `perpty` than yes I can put this in a Mixin.

There's a difference though, because here the design is stored in `.obsm["design_matrix"]` as opposed to `.design` in `pertpy`.

pydeseq2/dds.py
BorisMuzellec: Good suggestion, I created #329 to keep track of this.

pydeseq2/dds.py
BorisMuzellec: Opened #330 and #331 to keep track of this.

pydeseq2/dds.py
grst: I'm not entirely sure yet what's the best solution. Maybe we just leave it as is now, and when I'll look into refactoring pertpy I can propose a PR with changes to pyDESeq2 if required.

pydeseq2/ds.py
grst: Let's consider a design `~ disease * timepoint`

| disease | timepoint |
|--|--|
| healthy | T0 |
| healthy | T1 |
| diseased | T0 | 
| diseased | T1 | 

which gives us the following coefficients:
`Intercept`, `diseased`, `T1`, `T1:diseased`


Then you could test:

diseased vs. healthy
```python
contrast = dds.cond(disease="diseased") - dds.cond(disease="healthy")
```

T1 vs. T0
```python
contrast = dds.cond(timepoint="T1") - dds.cond(timepoint="T0")
```

Interaction T1:diseased
```python
contrast = (
    (dds.cond(timepoint="T1", disease="diseased") - dds.cond(timepoint="T0", disease="diseased")) - 
    (dds.cond(timepoint="T1", disease="healthy") - dds.cond(timepoint="T0", disease="healthy"))
)
```

grst: I admit there's no suitable documentation for this in pertpy. But in principle, using this "DSL", it should be possible to specify arbitrary contrasts.

pydeseq2/dds.py
```diff
@@ -226,55 +237,57 @@ def __init__(
...
+            )
+            design_factors = (
+                design_factors if isinstance(design_factors, list) else [design_factors]
+            )
+            self.design = "~" + " + ".join(design_factors)
 
-            # Also check continuous factors
-            if self.continuous_factors is not None:
-                self.continuous_factors = replace_underscores(self.continuous_factors)
+        assert isinstance(self.design, (str, pd.DataFrame)) or isinstance(
```
grst, line 274: nitpick: if this is meant as a user-facing error message, it should probably be a ValueError instead of an assertion

grst (approved):
LGTM

umarteauowkin (approved):
Thanks  for this great PR, I m convinced :) Just one thing that is not clear for me is what should be put in the LFC shrinkage: is it really a column of the design ? For me it should be LFC @ contrast, but maybe this is not relavant for this PR, I just spotted it since you made the modification. 
Finally, could you add a comment on what the Materializer is supposed to do ? (i.e., just one line of comment on what a materializer is :))
Thanks again !

BorisMuzellec:
>  Just one thing that is not clear for me is what should be put in the LFC shrinkage: is it really a column of the design ?

Yes: LFC shrinkage performs MAP estimation with a prior on a given LFC coefficient (i.e. column) that the user must specify.

In principle, I guess it would be possible to do the same thing with contrast @ LFC, I'm just not sure what it would mean. (Also, would the prior be amenable to linear combination?)

BorisMuzellec:
Thanks for your reviews @grst @umarteauowkin ! I'm merging this :)

PR #332, opened 2024-11-04
MAINT Deprecated `pandas.DataFrameGroupBy.grouper`

BorisMuzellec:

`FutureWarning`s were ignored in the CI, which is not good practice. They are no longer ignored by default.

`pandas.DataFrameGroupBy.grouper` is now deprecated. This PR replaces it with a call to `ngroup()`. 

It also handles a warning due to changes in `rcond` default argument value in `np.linalg.lstsq()`.

Two `Futurewarning`s were specifically ignored, as they will potentially no longer be relevant after #328 is merged.


umarteauowkin (approved):
LGTM thanks :)

PR #344, opened 2024-12-01
Switch to formulaic-contrasts

grst:
Hi @BorisMuzellec, 

I created [formulaic-contrasts](https://github.com/scverse/formulaic-contrasts) containing all the code required for introspecting formulaic models and defining contrasts. 

As described [here](https://formulaic-contrasts.readthedocs.io/en/latest/model_usage.html) it can be used either by inheritance or composition. Using inheritance would save us from redefining `.variables`, `.cond` and `.contrast`, but there'd be a name clash with the `design` variable so I went for the composition approach instead. Using inheritance/mixin would be possible with a little bit of refactoring, but I'd leave that to you should you prefer this approach. 

There's now also documentation on how to build contrasts: https://formulaic-contrasts.readthedocs.io/en/latest/contrasts.html 
Feel free to reference that from and/or copy it to the PyDESeq2 docs. 

Best, 
Gregor


BorisMuzellec:
Hi @grst,

Thanks for creating this PR! I'm happy with the way you propose to reorganize things :)

I'll add a reference to https://formulaic-contrasts.readthedocs.io/en/latest/contrasts.html in the docs when I'll have some time, let me open an issue to remember to do it

BorisMuzellec (approved):

tests/test_pydeseq2.py
```diff
@@ -146,7 +146,7 @@ def test_deseq_independent_filtering_mean_fit(counts_df, metadata, tol=0.02):
     )
     dds.deseq2()
 
-    ds = DeseqStats(dds, contrast=["condition", "A", "B"])
+    ds = DeseqStats(dds, contrast=["condition", "B", "A"])
```
grst, line 149: hang on, is this what you expect? it should be [variable, baseline, group_to_compare], right? I just want to make sure I didn't mix up things in formulaic-contrasts

BorisMuzellec: It should be `[variable, group_to_compare, baseline]`, and I think that you are right and that I got confused in #328 

It used to be optional to input a contrast to `DeseqStats` before #328. I went a bit too fast in #328 and put the contrasts that passed the tests, but I realised that the code used to be 

```python
    def _contrast(self, column: str, baseline: str, group_to_compare: str) -> np.ndarray:
        """Build a simple contrast for pairwise comparisons.
        This is equivalent to
        ```
        model.cond(<column> = baseline) - model.cond(<column> = group_to_compare)
        ```
        Parameters
        ----------
        column: str
            The column to contrast.
        baseline: str
            The baseline group.
        group_to_compare: str
            The group to compare to the baseline.
        Returns
        -------
        np.ndarray
            The contrast vector.
        """
        return self.dds.cond(**{column: baseline}) - self.dds.cond(
            **{column: group_to_compare}
        )
```

Which (unless I'm mistaken again!) was wrong.

BorisMuzellec: Which is indeed what is implemented in pertpy: https://github.com/scverse/pertpy/blob/0091d0dcb6139449a5a67a1255635afd0e9e4cfb/pertpy/tools/_differential_gene_expression/_base.py#L1089

Issue #320, opened 2024-09-27
Feature Request: RLE Plots


jonathjd:
**Is your feature request related to a problem? Please describe.**

Right now there is no method that plots an RLE plot even though the `DeseqDataSet` method has both the normalized counts (`DeseqDataSet.layers["norm_counts"]`) and estimated size factors (`DeseqDataSet.obsm["size_factors"]`). RLE plots are useful for identifying technical variation and normalization issues in RNA-Seq data, as well as testing out different normalization strategies. These plots are super helpful in assessing and visualizing unwanted technical noise or batch effects in expression data.

**Describe the solution you'd like**
I propose implementing an RLE plot method, similar to the `plotRLE` function in EDASeq for R. The plot would display boxplots of the relative log expression of genes for each sample, centered around the median per gene. The implementation would:

- Compute the median for each gene across all samples.
- Calculate the log expression values relative to this median.
- Generate a boxplot for each sample based on relative log expression values.

**Describe alternatives you've considered**

The implementation in other standard libraries (pandas, numpy, matplotlib) is not too bad, but it would be super handy if it came standard!

**Additional context**
Here's an example of an RLE plot I've made that could be similar. Thanks!!

<img width="493" alt="Screenshot 2024-09-27 at 4 38 23 PM" src="https://github.com/user-attachments/assets/5ed94f44-dac3-4996-9714-7b89de2d1b7e">

BorisMuzellec:
Hi @jonathjd, great suggestion! 

I probably won't be able to implement this myself, but I'd love to help you open a PR.

This could be a method of the `DeseqDataSet` class wrapping a function in `utils.py`, similarly to [`plot_dispersions`](https://github.com/owkin/PyDESeq2/blob/ab0db4babefb3958d89c410d17d9f14af4620ff9/pydeseq2/dds.py#L978) in `DeseqDataSet` which calls [`make_scatter`](https://github.com/owkin/PyDESeq2/blob/ab0db4babefb3958d89c410d17d9f14af4620ff9/pydeseq2/utils.py#L1398) from `utils`. 

Let me know if you need any help :).

jonathjd:
@BorisMuzellec That sounds like a great idea, and I would be happy to work on it! Let me know what I need to do to make it happen.

PR #348, opened 2024-12-07
Add Relative Log Expression (RLE) Plots

jonathjd:
#### Reference Issue or PRs

Issue Reference: https://github.com/owkin/PyDESeq2/issues/320

The Relative Log Expression (RLE) is a useful diagnostic plot to visualize the differences in count distributions between samples. The x-axis is each sample from the count matrix and the y-axis is the log difference between each gene and the median expression of that gene across all samples.

Given:

`gene_ij` is the expression of gene `j` in sample `i`
`median_j` is the median expression of gene `j` across all samples
The RLE for gene `j` in sample `i` is calculated as:

```
RLE_ij = Log2(gene_ij/median_j)
```

Where:

`gene_{ij} `is the count of gene `j` in sample `i`
`median_j` is the median count of gene j across all samples

This issue takes in the raw counts `self.X`, a `normalize` boolean, `design_matrix.index` for the sample_ids, and a `save_path` and produces an RLE plot. 

rle_plot

The normalize boolean is set to `False` by default but can be set to `True` to normalize the raw counts before plotting

nlog_rle_plot

This example was produced with the synthetic data in the `./datasets/` dir

### DIFF
--- a/pydeseq2/dds.py
+++ b/pydeseq2/dds.py
@@ -21,6 +21,7 @@ from pydeseq2.preprocessing import deseq2_norm_fit
 from pydeseq2.preprocessing import deseq2_norm_transform
 from pydeseq2.utils import build_design_matrix
 from pydeseq2.utils import dispersion_trend
+from pydeseq2.utils import make_rle_plot
 from pydeseq2.utils import make_scatter
 from pydeseq2.utils import mean_absolute_deviation
 from pydeseq2.utils import n_or_more_replicates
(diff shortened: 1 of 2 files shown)


pydeseq2/utils.py
```diff
@@ -1608,3 +1608,64 @@ def lowess(
...
+        (default: ``None``).
+
+    **kwargs :
+        Additional keyword arguments passed to matplotlib's boxplot function.
+    """
+    if normalize:
+        print("Plotting normalized RLE plot...")
+        geometric_mean = np.exp(np.mean(np.log(count_matrix + 1), axis=0))
+        size_factors = np.median(count_matrix / geometric_mean, axis=1)
+        count_matrix = count_matrix / size_factors[:, np.newaxis]
```
BorisMuzellec, line 1646: This is recomputing median-of-ratios size factors, right? (up to the +1 in the log)

In that case, it might be better to add a `size_factors` argument to `make_rle_plot`. 

From there, in `dds.plot_rle`, if `normalize=True`, we would first check whether size factors (`self.obsm["size_factors"]`) were already computed. If so, we pass them directly. If not, we call `self.fit_size_factors` first.

pydeseq2/dds.py
```diff
@@ -1010,6 +1011,36 @@ def plot_dispersions(
...
+
+        **kwargs
+            Keyword arguments for the scatter plot.
+        """
+        make_rle_plot(
+            count_matrix=self.X,
+            normalize=normalize,
+            sample_ids=self.obsm["design_matrix"].index,
+            save_path=save_path,
+            **kwargs,
```
BorisMuzellec, line 1041: c.f. my comment in `utils.make_rle_plot`: if normalize, check whether we already have size factors in `self.obsm["size_factors")`

BorisMuzellec (changes requested):
Thanks a lot for this PR @jonathjd !

I just have one suggestion to make: unless I missed something,I think we should avoid re-computing size factors when we already have them.

Happy to merge once this is fixed!

pydeseq2/utils.py
jonathjd: Hi @BorisMuzellec ,

For some reason when I use the size factors computed in dds.fit_size_factors() I do not get an RLE plot with the sample medians centered around 0.

image

But when I compute the sizefactors internally I do get an RLE plot with the sample medians centered around 0

image

Do you know what could be causing this?

jonathjd: @BorisMuzellec Happy holidays! Sorry for the delay, I was on vacation. It's probably a good idea to add the psuedocount to the `gene_median` calculations so we don't divide by a gene with 0 counts across all samples. I can add that now.

PR #350, opened 2024-12-12
changed @ to dot

avihani:

Instead of using '@' for matrix multiplication I wanted to use the pandas function pd.dot.
This effectively changes nothing in the underlying functionality or usage of the package.

However I suppress the syntactic sugar of @ with pandas so that I can use '@' for other purposes.
If the choice of using syntactic sugar is by design then it is understandable why this PR may be rejected.

### DIFF
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -305,7 +305,7 @@ class DeseqStats:
                 )
 
         mu = (
-            np.exp(self.design_matrix @ self.LFC.T)
+            np.exp(self.design_matrix.dot(self.LFC.T))
             .multiply(self.dds.obsm["size_factors"], 0)
             .values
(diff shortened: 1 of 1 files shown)


BorisMuzellec:
Hi @avihani ! Could you provide some insight on what you would use "@" for, and why using it in the codebase prevents it?

avihani:
I use '@' for custom defined functions mainly surrounding dataframe merging.
I prefer the .dot function overall more than the '@' for clarity though.
Similar to next line (after the one I modified) uses pd.mul instead of '*'

BorisMuzellec:
Sorry but I think that using the `@` syntactic sugar makes the codebase much more readable. I hope you find a way to work around this in your code 🙏

PR #379, opened 2025-04-07
ENH DeseqStats n_cpus

asistradition:

Adds n_cpus argument to DeseqStats with the same logic as DeseqDataSet


BorisMuzellec (approved):
LGMT, thanks!

PR #412, opened 2025-09-13
Normalization factors and DESeqDataSetFromTximport equivalent

### LINKED ISSUE #305: Add support for sample-/gene-dependent normalization factors (e.g., length offsets from pytximport)
## Request

I would like to propose support for sample-/gene-dependent normalization factors to enable the use length offset matrices from [pytximport](https://pytximport.readthedocs.io/en/latest/start.html), a full-featured Python implementation of tximport, similar to how DESeq2 supports length matrices from tximport. With this issue, I would like to see if there is interest in this addition and discuss how best to implement it given the current design of PyDESeq2.

## Context

DESeq2 implements the `DESeqDataSetFromTximport` function, that takes a tximport object and creates a DESeqDataSet from it that utilises the abundance-weighted average gene length across isoforms for each sample to calculate normalisation factors. The rationale is that differential isoform usage may lead to different average gene lengths between samples and introduce a bias that needs to be corrected for. See also: https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/vignettes/DESeq2.Rmd#L2181

DESeq2 uses this sample-specific average gene length when `DESeqDataSet.estimateSizeFactors` is called. In short, the sample-specific average gene length from tximport is divided by the gene-wise geometric mean to create a normalization matrix which is then passed together with the counts to `estimateNormFactors`, where the counts divided by the normalization matrix are passed to `estimateSizeFactorsForMatrix` to calculate size factors. Then the normalization factors are calculated: `norm_factors = (normMatrix.T*size_factors).T / geometric_mean((normMatrix.T*size_factors).T)`. These normalization factors are then used instead of the size factors for various downstream tasks.

Relevant parts of the DESeq2 code:
https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/R/AllClasses.R#L408
https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/R/methods.R#L383
https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/R/core.R#L2185
https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/R/core.R#L542

## Challenges for implementation

With the current design of PyDESeq2, it might not be necessary to implement a standalone function like `DESeqDataSetFromTximport`, since that function only adds a key to the DDS. When using the PyDESeq2 DDS with an AnnData object created by pytximport, the "length" observation matrix and the unstructured key “counts_from_abundance” will already be present, eliminating the need for a DESeqDataSetFromTximport function. A check whether a length offset matrix can be used is as simple as:
```python
if "length" in self.obsm_keys() and "counts_from_abundance" in self.uns_keys() and self.uns["counts_from_abundance"] is None:
    # do something
```

However, the implementation of the size factor fitting and count normalization currently differs significantly between the R and Python version of DESeq2, making it non-trivial to implement support for normalization factors. As far as I understand the DESeq2 and PyDESeq2 code, the following steps would be necessary to add support for normalization factors based on length offsets from pytximport.

When `fit_type`is `ratio`, `DeSeqDataset.fit_size_factors` would have to detect the presence of a precomputed normalization matrix or of a length matrix that could be transformed into a normalization matrix.
Then, like with DESeq2, there would be two options:
- No normalization/length matrix => call the equivalent to `estimateSizeFactorsForMatrix` (`deseq2_norm_fit` & `deseq2_norm_transform` in the case of PyDESeq2, current implementation)
- Normalization matrix present => implement an equivalent of `estimateNormFactors`, for example as part of `DeSeqDataset.fit_size_factors`

Pseudocode implementation:
```python
def fit_size_factors(..., norm_matrix=None,) -> None:

    # ...

    if "length" in self.obsm_keys() and "counts_from_abundance" in self.uns_keys() and self.uns["counts_from_abundance"] is None:
        norm_matrix = self.obsm["length"] / row_wise_geometric_mean(self.obsm["length"])

    if norm_matrix:
        counts = self.X / norm_matrix
        log_means, filtered_genes = deseq2_norm_fit(counts)
        _, size_factors = deseq2_norm_transform(counts, self.logmeans)
        normalization_factors = (norm_matrix.T*size_factors).T
        normalization_factors = normalization_factors / row_wise_geometric_mean(normalization_factors)
    else:
        # no change
```

Then, support of the normalization factors in the downstream methods would have to be implemented. `DESeq2` uses them as part of the negative binominal GLM fitting used with the Wald test, likelihood ratio test and LFC shrinking. See:
- Wald test: https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/R/fitNbinomGLMs.R#L61
- Likelihood ratio test: https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/R/core.R#L1935
- LFC shrinking: https://github.com/thelovelab/DESeq2/blob/6e1a3a56f7f6315e4a5ff352a203af26e778f73a/R/lfcShrink.R#L362

Implementing this would therefore affect several parts of the PyDESeq2 codebase, so I am opening this issue to discuss the need and best way to implement these changes first. In my opinion, support for normalization factors would be an important addition to PyDESeq2, not only to support correction for differential isoform usage based on the output of pytximport, but also to allow corrections for differences in gene length for multi-laboratory or multi-timepoint datasets. I am happy to contribute code to implement this feature and look forward to your feedback.

As the maintainer of pytximport, I can also change naming conventions or add further metadata to the output of pytximport AnnData objects to facilitate integration with PyDESeq2, if you think that relying on the presence of the `counts_from_abundance` key in the unstructured part of the AnnData object is not sufficient.

Best,
Malte

### LINKED ISSUE #359: Plans to add DESeqDataSetFromTximport?
Hi, 
I was wondering if there is any plans to add a functionality similar to R's DESeqDataSetFromTximport()?
Currently salmon data cannot be used in python unless it is rounded, which is not really the correct way to do it.
Since there now is also a pytximport package I was wondering if something is in the works.
Best
Anne

maltekuehl:
Fixes #305 and closes #359. Adds support for normalization factors based on length provided by `pytximport`. Very much still a draft.

Open issues:
- Is the example dataset okay? Others seem to be synthetic and much smaller. Do you have some simple or synthetic source data that could be used to create an AnnData object with pytximport? Once clear, we should probably also add tests for correctness, comparing against a reference to prevent accidental future drift.
- I have yet to figure out how size factors should be implemented in the case that we are also calculating normalization factors. Should they be equivalent to size factors in the standard case or some transform of the norm matrix adjusted counts? Similar for logmeans, as I do not fully understand where else this data is used throughout the code. Any help would be appreciated.
- The statistical testing structure is quite a bit different in PyDESeq2 compared to DESeq2 and I may not have grasped all subtleties, would be thankful for help from maintainers to ensure that everything was adjusted.
- The initial scaffold of this PR was LLM-generated, and while I provided ample context (including the issues and relevant code from DESeq2) and clear instructions and have checked and already adjusted the output quite a bit, it would be best to examine these changes critically.

Aside: Tests pass locally but fail due to an older AnnData version on Python 3.10 here. Would you be open for this PR to also include an update of the pre-commit, GitHub Actions, full move to uv/hatch/ruff like other scverse ecosystem packages and targeting Python 3.11 - 3.13?

CC @BorisMuzellec


gitbenlewis:
Hi @maltekuehl, thank you for working on this. I encountered the same need and only found this draft after implementing a tested version in my fork.

I have described the implementation in [issue #305](https://github.com/scverse/PyDESeq2/issues/305), and the code is available here:

https://github.com/gitbenlewis/PyDESeq2/commit/3e9eb78a6d856d7744f092e8b6b0841a1ae85682

My implementation uses an optional `transcript_lengths` argument and stores the inputs and fitted normalization factors in AnnData layers. It includes targeted coverage for the normalization formula, `ratio` and `poscounts`, input validation, sparse AnnData input, dispersion fitting, Wald testing, LFC shrinkage, Cook’s outlier replacement and refitting, and VST.

I do not want to duplicate or compete with your work. Would you prefer that I contribute relevant changes and tests to this PR, or open a separate draft PR so the maintainers can compare the approaches?

I would be happy to coordinate and adapt the API or implementation based on your and the maintainers’ feedback.

maltekuehl:
Hi @gitbenlewis, great to see that there is interest in this!

I unfortunately had to abandon this PR a while ago due to lack of time, but I think it should still be relevant and the approach had been discussed with @BorisMuzellec beforehand.

I am open to somebody taking over this draft PR and driving the implementation of the change. If you have the availability and would be willing to contribute, from my side that's great, though the project maintainers would also need to comment (cc, @Zethson).

I would however encourage you to explore the approach I was taking beforehand and to create a comparison with your approach to discuss this together between everybody before you go on to create a full PR.

One important aspect that this PR wanted to cover is the connection with https://github.com/complextissue/pytximport which would be the natural fit in any Python workflow, so whichever way this is implemented, it would be great if that compatibility could be kept and the examples and docs could point there, too (while of course still crediting the intellectual work done by the original tximport team!).

Zethson:
> If you have the availability and would be willing to contribute, from my side that's great, though the project maintainers would also need to comment (cc, @Zethson).

I'm more than happy if someone steps in and finishes some PRs. Thanks @gitbenlewis !

gitbenlewis:
Thanks again, @maltekuehl, for the guidance and for being open to someone taking this work forward. Thanks also to @Zethson for the encouragement.

  After comparing the approaches, I extended my implementation with direct compatibility for the current `pytximport` AnnData schema and opened #451.

  The new PR preserves the generic `transcript_lengths` API while also accepting `pytximport` output through `obsm["length"]` and `uns["counts_from_abundance"]`. Both routes use the same normalization-factor implementation. It rejects abundance-scaled count modes, preserves sparse CSR/CSC counts, and does not add `pytximport` as a runtime dependency.

  The tests cover normalization, dispersion and LFC fitting, Wald tests, shrinkage, VST, and Cook’s outlier replacement/refitting. I have credited the  earlier work in #412, as well as the original `pytximport`, tximport, and DESeq2 work, in the PR and documentation.

  I would especially appreciate any corrections regarding the `pytximport` schema or intended workflow. If the maintainers agree, #451 can serve as a ready-for-review continuation of this draft.

PR #416, opened 2025-09-24
Build with uv, hatchling, updated CI/pre-commit and SPEC 0 compliant package versions

maltekuehl:
#### Reference Issue or PRs
Discussed with @BorisMuzellec in #412.

- Uses `uv` for package management and `hatchling` for building, similar to most `scverse` core packages.
- Updates dependencies to be in line with [Scientific Python SPEC 0](https://scientific-python.org/specs/spec-0000/)
- Additional checks for `pre-commit`
- Fixes many `mypy` errors that previously went undetected (this is the reason that many files are affected in this PR)
- Updates the CI pipeline to run the `ruff` and `mypy` checks, too, and to test building the package
- Removes `black` as it has been superseded by `ruff` and the old `setup.py` installation, which now works through `pyproject.toml`
- Updates documentation to include installation with `uv`
- Increments the version by a patch version, as no public-facing API should have been affected

There is some debate about whether to commit `uv.lock` files, [Astral's position](https://github.com/astral-sh/uv/issues/9797) is YES for full reproducibility, but if you would prefer this to not be committed, please let me know. Downside is that if this is committed, the CI does not test against the latest version of all dependencies but against the locked one, unless `--upgrade` is passed. Passing this option would also be a possible solution to this question.

### DIFF
--- a/pydeseq2/ds.py
+++ b/pydeseq2/ds.py
@@ -511,7 +511,7 @@ class DeseqStats:
             U2 = self.p_values[use]
             if not U2.empty:
                 result.loc[use, i] = false_discovery_control(U2, method="bh")
-        num_rej = (result < self.alpha).sum(0).to_numpy()
+        num_rej = (result < self.alpha).sum(0).to_numpy().astype(int)
         lowess_res = lowess(theta, num_rej, frac=1 / 5)
 
         if num_rej.max() <= 10:


maltekuehl:
Fixed! Ready to merge from my side now, please let me know if there's anything I missed.

PR #441, opened 2026-02-25
jax deseq2

adamgayoso:
Implement a `JaxInference` backend for DESeq2 inference.

This backend allows hardware acceleration via GPU/TPU. Please see the class docstring for a detailed description of the methodology.

A test file has been added to test all the same use cases as the default inference class and all tests pass.

Based on the import structure, the class can only be imported if jax is installed, so no added logic was implemented for handling missing imports.

### DIFF
--- a/pydeseq2/jax_inference.py
+++ b/pydeseq2/jax_inference.py
@@ -1,0 +1,1341 @@
+"""PyDESeq2 inference with a jax backend."""
+
+import functools
+from collections.abc import Callable
+from typing import Any
+from typing import Literal
+
+import chex
+import jax
+import jax.numpy as jnp
+import jax.scipy
+import numpy as np
+import optax
+import pandas as pd
+
+from pydeseq2 import inference
+
+
+# Jax lstsq can be inaccurate https://github.com/google/jax/issues/11433
+def _lstsq(a, b):
+    return jnp.linalg.solve(a.T @ a, a.T @ b)
+
+
+def _maybe_cast_to_float64(x):
+    return x.astype(jnp.float64) if jax.config.x64_enabled else x
+
+
+def _optax_solver(
+    init_params: chex.Array,
+    fun: Callable[..., chex.Array],
+    opt: optax.GradientTransformationExtraArgs,
+    max_iter: int,
+    g_tol: float = 1e-5,
+) -> tuple[chex.Array, chex.Array]:
+    """Interface to minimize using optax."""
+    value_and_grad_fun = optax.value_and_grad_from_state(fun)
+
+    def step(carry):
+        params, state = carry
+        value, grad = value_and_grad_fun(params, state=state)
+        updates, state = opt.update(
+            _maybe_cast_to_float64(grad),
+            state,
+            params,
+            value=value,
+            grad=grad,
+            value_fn=fun,
+        )
+        params = optax.apply_updates(params, updates)
+        return params, state
+
+    def continuing_criterion(carry):
+        _, state = carry
+        iter_num = optax.tree_utils.tree_get(state, "count")
+        grad = optax.tree_utils.tree_get(state, "grad")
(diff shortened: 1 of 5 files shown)


pydeseq2/jax_inference.py
```diff
@@ -0,0 +1,1341 @@
...
+    )
+    beta_, mu_, converged_ = jax.vmap(
+        _irls_loop,
+        in_axes=(1, None),
+        out_axes=(1, 1, 0),
+    )(gene_args, global_args)
+    return beta_, mu_, converged_
+
+
+def _compute_hat_matrix_diagonal(
```
ilan-gold, line 442: array-api-able? https://github.com/scverse/PyDESeq2/blob/main/pydeseq2/utils.py#L427-L433 + https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html

pydeseq2/jax_inference.py
```diff
@@ -0,0 +1,1341 @@
...
+    Returns
+    -------
+    Negative log likelihood of the observations counts.
+    """
+    return _nb_loss_constant_terms(counts, dispersion) + _nb_loss_variable_terms(
+        counts, mean, dispersion
+    )
+
+
+def _compute_mu(
```
ilan-gold, line 140: array-api-able? https://data-apis.org/array-api/latest/API_specification/generated/array_api.maximum.html#maximum

pydeseq2/jax_inference.py
```diff
@@ -0,0 +1,1341 @@
...
+) -> jnp.ndarray:
+    """Loss to minimize for fit alpha mle."""
+    # A leading dimension is necessary for optimization.
+    log_alpha = log_alpha[0]
+    alpha = jnp.exp(log_alpha)
+
+    w = mu / (1 + mu * alpha)
+    mat = (design_matrix.T * w) @ design_matrix
+
+    # There are a few ways to compute the log determinant. Here we use an approach
```
ilan-gold, line 547: Why not https://docs.jax.dev/en/latest/_autosummary/jax.numpy.linalg.slogdet.html? It seems like it's in the array-api: https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.slogdet.html#slogdet and also curious if it has the same issue as the torch one?

pydeseq2/jax_inference.py
```diff
@@ -0,0 +1,1341 @@
...
+    dispersion_neg1 = 1 / dispersion
+    logbinom = (
+        jax.scipy.special.gammaln(counts + dispersion_neg1)
+        - jax.scipy.special.gammaln(counts + 1)
+        - jax.scipy.special.gammaln(dispersion_neg1)
+    )
+    return n * dispersion_neg1 * jnp.log(dispersion) - logbinom.sum()
+
+
+def _nb_loss_variable_terms(
```
ilan-gold, line 103: array-api-able? https://data-apis.org/array-api/latest/API_specification/generated/array_api.log.html#log + part of https://github.com/scverse/PyDESeq2/blob/main/pydeseq2/utils.py#L163

ilan-gold (commented):
Spent some time with this, seems like I would have some broad questions:

1. Is it worth it to abstract the "math" from the "jax" via `array-api`? My comments are along these lines. My software engineering instincts are "yes" but I don't know Jax or the math well enough to say either way and also it's a massive lift as it would require unifying with the current implementation.  I'll try to spend some time reading up
2. In general, I see a few uses of things that have multiple implementations in `jax`, specifically where one implementation lives in `jax.spicy` and the other in `jax.numpy` - It seems there are good reasons for this usually like slightly different features. But that raises the question further from the the above of how much of this could be refactored to be reused (modulo which `scipy` module is called for a `numpy` or `jax` array).

Not sure I'm the one most qualified to review, but figured I'd get the ball rolling! Maybe we should discuss among core what to do.

